PR #1, opened 2018-05-17
Disable route by default and enable TLS

srueg:


prometheus/templates/server-route.yaml
```diff
@@ -27,4 +27,9 @@ spec:
     name: {{ template "prometheus.server.fullname" . }}
     weight: 100
   wildcardPolicy: None
+  {{- if .Valuse.server.route.tls }}
```
tobru: Typo :hammer: "Valuse" vs "Values"

PR #6, opened 2018-07-18
Initial Keycloak chart

srueg:
Copied from upstream [1] with adaptions to run it on OpenShift.

[1] https://github.com/kubernetes/charts/tree/master/stable/keycloak


mhutter:
Can you walk me through what was changed from upstream tomorrow?

PR #17, opened 2018-12-17
WIP: feature/oauth proxy

anothertobi:


openshift-oauth-proxy/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+description: OpenShift oauth-proxy
+name: openshift-oauth-proxy
+version: 0.1.0
+maintainers:
```
srueg: Please add an `appVersion` field.

openshift-oauth-proxy/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ .Values.serviceaccount }}
```
srueg: I think the SA need an annotation: `serviceaccounts.openshift.io/oauth-redirectreference.primary`

openshift-oauth-proxy/templates/deployment.yaml
```diff
@@ -0,0 +1,40 @@
...
+            - -openshift-service-account={{ .Values.serviceaccount }}
+            - -cookie-secret-file=/etc/proxy/secrets/session_secret
+          {{- range $key, $value := .Values.extraArgs }}
+            - -{{ $key }}={{ $value }}
+          {{- end }}
+          resources:
+{{ toYaml .Values.resources | indent 12 }}
+          volumeMounts:
+          - mountPath: /etc/proxy/secrets
+            name: {{ template "openshift-oauth-proxy.name" $ }}
```
srueg: Where does this volume come from?

PR #22, opened 2019-02-13
HelmQA integration

spiogit:
The Travis script is enhanced to perform HelmQA checking on the charts. It runs in SaaS mode by default but can also run in Docker mode or be disabled completely. Checks could fail or just warn, right now they always fail.

The code is not yet great, e.g. this could be outsourced into a second script which could ship with HelmQA to be maintained centrally. But it would be good to get some feedback on functionality by chart maintainers to drive the next steps.

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -7,7 +7,7 @@ echo "----> Deploying to $REPO_URL"
 
 GIT_REPO="$(git config remote.origin.url)"
 
-helm_status=$(curl --max-time 10 $HELMQA_URL=$REPO_URL | jq -r .status)
+helm_status=$(curl --max-time 10 "$HELMQA_URL=$REPO_URL" | jq -r .status)
 
 echo "###################################### $helm_status ######################################"
 
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -7,18 +7,21 @@ echo "----> Deploying to $REPO_URL"
 
 GIT_REPO="$(git config remote.origin.url)"
 
-HELM_RESPONSE=$(curl --connect-timeout 5 "$HELMQA_URL=$GIT_REPO")
+HELMQA_RESPONSE=$(curl --connect-timeout 5 "$HELMQA_URL=$GIT_REPO")
 
-HELM_STATUS=$(echo "$HELM_RESPONSE" | jq -r .status)
-HELM_RESPONSE_CODE=$(echo "$HELM_RESPONSE" | jq -r .code)
-
-echo "-------------------> $HELM_STATUS"
-echo "-------------------> $HELM_RESPONSE_CODE"
-
-if [[ "$HELM_STATUS" == "fail" ]]; then
-	echo "HelmQA test failed. Check response"
-	echo "-------------> $HELM_RESPONSE"
-	exit 1
+if [[ $? != 28 ]]; then
+	TEST_STATUS=$(echo "$HELMQA_RESPONSE" | jq -r .status)
+        echo "-------------------> $TEST_STATUS"
+	
+	if [[ "$TEST_STATUS" == "fail" ]]; then
+		echo "HelmQA test failed. Check response"
+		echo "-------------> $HELMQA_RESPONSE"
+		exit 1
+	else 
+		echo "HelmQA test succeeded. No issues found"
+	fi
+else
+	echo "HelmQA connection failed. Timed out!"
 fi
 
 tmp="$(mktemp -d)"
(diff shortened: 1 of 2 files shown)


spiogit:
Btw. the build failure is by design. In maxscale/Chart.yaml there is no maintainer information, which is clearly a quality issue.

.ci/build.sh
```diff
@@ -1,11 +1,48 @@
 #!/bin/bash
 set -e
 
+# set to one of the following values
+# - don't use HelmQA: (empty)
+# - use HelmQA in SaaS mode: http://helmqa-zhaw-prod-demos.appuioapp.ch/livecheck?repo
+# - use HelmQA locally in Docker mode: http://127.0.0.1:5000/livecheck?repo
+HELMQA_URL=http://helmqa-zhaw-prod-demos.appuioapp.ch/livecheck?repo
```
mhutter: Quote URLs in single-quotes (`'`) to avoid globbing

.ci/build.sh
```diff
@@ -1,11 +1,48 @@
...
 REPO_URL="${REPO_URL:-https://charts.appuio.ch}"
 echo "----> Deploying to $REPO_URL"
 
 GIT_REPO="$(git config remote.origin.url)"
 
+if [ ! -z $HELMQA_URL ]; then
```
mhutter: Use `-n` instead of `! -z`. Quote variables to avoid globbing and splitting. (eg. `"$HELMQA_URL"`, same goes for all variables)

.ci/build.sh
```diff
@@ -1,11 +1,48 @@
...
+  echo $HELMQA_URL | grep -q 127.0.0.1
+  did=
+  if [ $? == 0 ]; then
+    docker pull jszhaw/helmqa
+    did=`docker run -d -p 127.0.0.1:5000:5000 helmqa`
```
mhutter: Use `$()` instead, and put double quotes around it

.ci/build.sh
spiogit: done

.ci/build.sh
spiogit: done

.ci/build.sh
spiogit: done

spiogit:
Now the HelmQA status message is properly shown.

.ci/build.sh
```diff
@@ -1,11 +1,53 @@
...
 REPO_URL="${REPO_URL:-https://charts.appuio.ch}"
 echo "----> Deploying to $REPO_URL"
 
 GIT_REPO="$(git config remote.origin.url)"
 
+if [ -n "$HELMQA_URL" ]; then
+  did=
+  if [[ "$HELMQA_URL" =~ 127.0.0.1 ]]; then
+    docker pull jszhaw/helmqa
+    did="$(docker run -d -p 127.0.0.1:5000:5000 helmqa)"
```
mhutter: Must be `jszhaw/helmqa` instead of `helmqa`.

I would also recommend to use `docker.io/jszhaw/helmqa` here and in the pull command above. This prevents issues with systems that have multiple docker repositories configured.

.ci/build.sh
```diff
@@ -1,11 +1,53 @@
...
+  fi
+
+  while true; do
+    sleep 1
+    tresponse="$(curl --connect-timeout 5 $HELMQA_URL=$GIT_REPO)" || true
```
mhutter: I would `--max-time` instead of `--connect-timeout`. max-time sets a timeout for ALL parts of the request (connect, sending request, awaiting response, receiving response) while connect-timeout only sets a timeout for the connect phase.

.ci/build.sh
spiogit: done

.ci/build.sh
spiogit: These two flags cover different risks. We need connect-timeout with a relatively low value to protect against e.g. firewall DROP rules. On the contrary, max-time needs to consider potentially longer checks. We do not have much data on how much time checks take. I have therefore added max-time additionally with a reasonable default (60s), just to prevent jobs hanging for longer than max HTTP/TCP timeouts. We need some observation to tune all values.

spiogit:
Just as a note: This pull request is merely the integration of HelmQA into the pipeline. Additional work will be needed to fine-tune all checks and see that in day-to-day chart maintenance, they are (1) helpful/instructive, (2) covering more and more quality issues, but also (3) not in the way of getting things done by being too pedantic.

spiogit:
Even with the Charts.yml improvement, the build would fail after the merge because there are many duplicates. The duplicates check may however be full of false positives. Therefore I've oc-imported a modified version of HelmQA which no longer leads to failure due to duplicates. Upon the next build in the charts repo, it should show success now.

spiogit:
Now it shows success upon a forced reinvocation of the build, but strictly speaking this is because it did not find any new charts and therefore did not scan. It would presumably still be successful when charts change. We have to wait to really the results in an unprovoked live situation.

PR #23, opened 2019-03-04
Helm keycloak chart with additional route

janPhil:


keycloak/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: keycloak
-version: 3.1.0
+version: 3.1.1
```
srueg: I'd suggest a major version bump (4.x) for this

keycloak/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - name: keycloak
+    version: 4.4.0
+    repository: https://kubernetes-charts.storage.googleapis.com/
```
srueg: Please add a newline at the end of the file

keycloak/README.md
```diff
@@ -2,12 +2,18 @@
```
srueg: Please rewrite this REDME to link the upstream, instead of copying it. Also make sure to mention how to overwrite values from the upstream chart (ì.e. prepend `keycloak`).

keycloak/requirements.yaml
```diff
@@ -0,0 +1,4 @@
+dependencies:
+  - name: keycloak
+    version: 4.4.0
```
srueg: Apparently version `4.5.0` of the chart was released.

keycloak/Chart.yaml
srueg: Also, please update the `appVersion` field accordingly.

PR #28, opened 2019-03-15
Test for version bump

srueg:
If a chart was changed without having it's version bumped, fail the
build. In order for this to work, all chart versions were bumped by a
patch version.

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -58,8 +58,8 @@ trap cleanup 0 1 2 3 6 15
 
 charts=$(find ./* -maxdepth 1 -name Chart.yaml -exec dirname "{}" \;)
 for chart in $charts; do
-  echo "----> Update dependencies for ${chart}"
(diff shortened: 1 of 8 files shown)


.ci/build.sh
```diff
@@ -74,14 +74,24 @@ echo "----> Check out gh-pages branch"
...
 
 echo "----> Moving new charts into repo"
-mv -nv "${tmp}"/*.tgz "${OUT_DIR}/"
+mv -v "${tmp}"/*.tgz "${OUT_DIR}/"
 
 pushd "$OUT_DIR"
 new_charts="$(git status --short | grep '^??' --count ||:)"
+modified_charts="$(git status --short | grep '^.M' --count ||:)"
+
+if [ "$modified_charts" -ne 0 ]; then
```
mhutter: If I read correctly, this will only evaluate to `true` when charts have been modified?

PR #31, opened 2019-03-25
Bump keycloak

srueg:


.ci/build.sh
```diff
@@ -57,7 +57,25 @@ cleanup() {
...
 charts=$(find ./* -maxdepth 1 -name Chart.yaml -exec dirname "{}" \;)
+changed_files=$(git diff --name-only $TRAVIS_COMMIT_RANGE)
+
+echo "Changed files: $changed_files"
+
 for chart in $charts; do
+  chart=${chart:2}
+  chart_changed=false
+  for file in $changed_files; do
+    if [[ $file == "$chart/"* ]]; then
```
mhutter: Does this work if the first file in `$changed_files` does not belong to `$chart`?

srueg: It'll only be true if there is a changed file belonging to this chart.

PR #34, opened 2019-04-10
WIP: Stardog backup to S3

srueg:
There's no need to customize the service type or port and neither to
specify multiple hosts for the ingress.


srueg:
#38

PR #36, opened 2019-04-12
Affinity

srueg:


stardog/values.yaml
```diff
@@ -83,3 +83,13 @@ zookeeper:
     # serverPasswords: secretPassword,
   metrics:
     enabled: true
+  affinity:
+    podAntiAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+        - labelSelector:
+            matchExpressions:
```
anothertobi: We could add `{{ .Release.Name }}` for the `release`label in order to allow multiple zookeepers in the same namespace.

srueg: Good catch but it's unfortunately not possible to use variables in the `values.yaml` file.

srueg: But I found another issue.

srueg: Ok, it's solved.

PR #37, opened 2019-04-17
[Stardog] various improvements

srueg:


ccremer:
lgtm
I have tested a `helm template` with and without zookeeper, but not actually rolled out on a k8s cluster. Is a more thorough testing (with failover etc.) planned for later before going into production?

PR #45, opened 2019-04-30
[stardog] Export metrics

srueg:


stardog/files/stardog.properties
```diff
@@ -1 +1,4 @@
 watchdog.enabled=false
+metrics.enabled=true
```
anothertobi: Is there a way to make this configurable? This might have a performance impact and not everyone will need metrics enabled.

stardog/files/jmx-exporter.yaml
```diff
@@ -0,0 +1,8 @@
+---
+startDelaySeconds: 10
+hostPort: 127.0.0.1:5833
+ssl: false
+lowercaseOutputName: true
+lowercaseOutputLabelNames: true
+rules:
```
anothertobi: Did you configure the defaults for the rules on purpose?

anothertobi:
LGTM

PR #46, opened 2019-05-07
[signalilo] Create Helm chart for Signalilo

srueg:
Resolves APPU-1564

### DIFF
--- a/signalilo/values.yaml
+++ b/signalilo/values.yaml
@@ -1,0 +1,50 @@
+# Default values for signalilo.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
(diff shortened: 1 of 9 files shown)


signalilo/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 2
+
+image:
+  registry: registry.vshn.net
+  repository: vshn/signalilo
+  tag: latest
+  pullPolicy: IfNotPresent
```
mhutter: Does not go well with `tag: latest`. Please change either

PR #49, opened 2019-05-14
[keycloak] Switch to new upstream chart

srueg:
The one in the stable repo was deprecated in favor of the new one in the
Codecentric repo.


keycloak/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: keycloak
-version: 4.8.0
+version: 4.9.0
```
anothertobi: We could keep up with the upstream chart version (4.12.0) as before

PR #55, opened 2019-05-16
Remove DCO from PR template

tobru:
We don't need this atm


srueg:
I'd propose to leave it and activate https://probot.github.io/apps/dco/ to enforce it

PR #58, opened 2019-05-31
[stardog] Set up monitoring and alerting

srueg:

Set up monitoring and alerts using Prometheus operator.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/stardog/values.yaml
+++ b/stardog/values.yaml
@@ -63,6 +63,7 @@ persistence:
 
 metrics:
   enabled: true
+  prometheusOperator: false
   image:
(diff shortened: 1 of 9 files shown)


stardog/values.yaml
```diff
@@ -63,6 +63,7 @@ persistence:
 
 metrics:
   enabled: true
+  prometheusOperator: true
```
anothertobi: Wouldn't enable it by default. The chart should be working with the default values and this won't be the case if there is an external dependency to the `prometheus-operator`.

PR #59, opened 2019-06-14
[stardog] Remove PVC label, upgrade

srueg:
Remove labels from PVC.
Upgrade Stardog.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/stardog/Chart.yaml
+++ b/stardog/Chart.yaml
@@ -1,8 +1,8 @@
 ---
 apiVersion: v1
 name: stardog
-version: 0.0.11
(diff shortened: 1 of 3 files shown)


srueg:
> Why are we removing the label from the PVC?

Because `volumeClaimTemplates` of StatefulSets are immutable and this label is changing on every new chart release. See https://github.com/helm/charts/issues/7803 for reference.

anothertobi:
Thanks for the clarification 👍 
Change LGTM

PR #60, opened 2019-06-20
[signalilo] Use environment variables for configuration

simu:
Signalilo v0.0.2 expects to be configured through environment variables.
This commit refactors the helm chart to provide the required config
through environment variables, from values provided in `.Values.config`
and allows extra configuration values to be provided as environment
variables in `.Values.extraEnvVars`.

Signed-off-by: Simon Gerber <simon.gerber@vshn.ch>

The new version v0.0.2 of Signalilo is configurable via environment variables instead of a config file. This PR refactors the Signalilo chart to provide the configuration via environment variables.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/signalilo/templates/deployment.yaml
+++ b/signalilo/templates/deployment.yaml
@@ -18,6 +18,8 @@ spec:
       labels:
         app.kubernetes.io/name: {{ include "signalilo.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
+      annotations:
+        checksum/tokens-secret: {{ include (print .Template.BasePath "/secret.yaml") . | sha256sum }}
     spec:
       {{- if .Values.securityContext.enabled }}
       securityContext:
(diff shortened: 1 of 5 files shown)


signalilo/templates/deployment.yaml
```diff
@@ -27,10 +29,31 @@ spec:
...
+          - name: SIGNALILO_ICINGA_USERNAME
+            value: {{ required ".Values.config.icinga_username is required" .Values.config.icinga_username | quote }}
+          - name: SIGNALILO_ICINGA_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: {{ include "signalilo.fullname" . }}
+                key: "icinga_password"
+          - name: SIGNALILO_ALERTMANAGER_PORT
+            value: {{ .Values.config.alertmanager_port | quote }}
+          - name: SIGNALILO_ALERTMANAGER_BEARER_TOKEN
```
srueg: This probably also needs to go into the secret.

mhutter:
It would be nice to use credentials from a preexisting secret. That way secrets can be deployed with sealedsecrets

signalilo/templates/deployment.yaml
simu: Moved to secret

simu:
> It would be nice to use credentials from a preexisting secret. That way secrets can be deployed with sealedsecrets

Modified chart so that both secret tokens can also be provided as pre-existing secret

signalilo/templates/deployment.yaml
```diff
@@ -27,10 +29,34 @@ spec:
...
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Values.config.alertmanager_bearer_token_secret | default (printf "%s" (include "signalilo.fullname" .)) | quote }}
+                key: "alertmanager_bearer_token"
+{{- if .Values.extraEnvVars }}
+{{ toYaml .Values.extraEnvVars | indent 10 }}
```
srueg: I think you can leave out the `toYaml` and just indent the block `.Values.k8up_operator.extraEnvVars`.

simu: unfortunately that doesn't work as expected and produces `[map[name:SIGNALILO_DEBUG value:3]]` instead of YAML

srueg: Oh, well. Never mind then.

PR #61, opened 2019-06-21
[stardog] Various improvements

srueg:
Various improvements.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/stardog/files/jmx-exporter.yaml
+++ b/stardog/files/jmx-exporter.yaml
@@ -9,6 +9,10 @@ rules:
     name: stardog_database_$2
     labels:
       database: $1
+  - pattern: Stardog<00=databases, 01=(.+?), name=(.+)><>(Count|Value)
+    name: stardog_database_$2
+    labels:
+      database: $1
   - pattern: Stardog<name=dbms\.mem\.(.+)><>Value
     name: stardog_dbms_mem_$1
   - pattern: .*
--- a/stardog/requirements.lock
+++ b/stardog/requirements.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: zookeeper
   repository: https://charts.bitnami.com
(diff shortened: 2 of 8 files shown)


stardog/values.yaml
```diff
@@ -22,6 +22,7 @@ stardog:
   existingLicenseSecret: stardog-license
   # Custom Stardog config (string value)
   config: |
+  javaArgs: -Xmx3g -Xms1g -XX:MaxDirectMemorySize=3g
```
anothertobi: Since `Xmx` is set, we should set the resource requests and limits as well.

stardog/files/jmx-exporter.yaml
```diff
@@ -9,6 +9,10 @@ rules:
     name: stardog_database_$2
     labels:
       database: $1
+  - pattern: Stardog<00=databases, 01=(.+?), name=(.+)><>(Count|Value)
+    name: stardog_database_$2
```
anothertobi: Is it possible to reuse the name?

stardog/requirements.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
 - name: zookeeper
   repository: https://charts.bitnami.com
-  version: 2.2.2
-digest: sha256:65da1c2229daa2c7da1465bb79115109c798f7477c6c26e0375de05dcca552d4
-generated: "2019-05-28T09:11:41.445617223+02:00"
+  version: 3.0.3
```
srueg: Apparently not compatible with the metrics exporter.

stardog/files/jmx-exporter.yaml
srueg: Yes, it shouldn't collide. There's an open ticket with Stardog to unify the metric names so we can get rid of this.

PR #64, opened 2019-06-26
[stardog] Manage stardog users with helm

janPhil:
#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/stardog/Chart.yaml
+++ b/stardog/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v1
 name: stardog
-version: 0.0.14
+version: 0.0.15
 appVersion: 6.2.2
 description: Stardog Helm Chart
 home: "https://www.stardog.com/"
--- a/stardog/README.md
+++ b/stardog/README.md
@@ -26,6 +26,8 @@ The following table lists the configurable parameters chart. For default values 
 | `image.existingPullSecret`                   | An existing image pull secret to use |
 | `stardog.existingLicenseSecret`              | Name of an existing secret with a key `stardog-license-key.bin` which contains the Stardog license |
 | `stardog.config`                             | Custom configuration to append to the `stardog.properties` file (as string) |
+| `stardog.adminPassword`                      | Password for the Administrator |
+| `stardog.users`                              | Array of users, each one with name and password |
 | `stardog.javaArgs`                           | Custom java args, passed to the JVM |
 | `stardog.securityContext`                    | K8s security context |
 | `stardog.backup.databases[]`                 | List of DBs to be backed up. The schedule is in UTC |
--- a/stardog/templates/post-install-manage-users.yaml
+++ b/stardog/templates/post-install-manage-users.yaml
@@ -1,0 +1,67 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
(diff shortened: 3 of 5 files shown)


srueg:
Please write in the pull request description what it is doing. Also make sure the check boxes are rendered correctly (remove the whitespace).

stardog/templates/post-install-add-users.yaml
```diff
@@ -0,0 +1,63 @@
...
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
+    "helm.sh/hook": post-upgrade,post-install
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/managed-by: {{.Release.Service | quote }}
```
srueg: Please use the same labels for the pod and the job as the StatefulSet does (i.e. `app.kubernetes.io/name` and `app.kubernetes.io/instance`).

stardog/templates/post-install-add-users.yaml
```diff
@@ -0,0 +1,63 @@
...
+            secretKeyRef:
+              name: {{ include "stardog.fullname" $ }}-user-passwords
+              key: {{ .name }}
+        {{ end }}
+        command:
+        - /bin/sh
+        - -ce
+        - |
+          echo "testing stardog connectivity to http://{{ include "stardog.fullname" $ }}:5820"
+          until $(curl -o /dev/null -s -f {{ include "stardog.fullname" . }}); do sleep 1 && echo "."; done
```
srueg: Please use port `5820` instead of `80` (on `80` it's the metrics).
I'd propose to create a env variable with the full server URL and to reuse it in all commands.

stardog/templates/post-install-add-users.yaml
```diff
@@ -0,0 +1,63 @@
...
+          {{- if .Release.IsInstall }}
+          echo Changing admin password
+          /opt/stardog/bin/stardog-admin --server http://{{ include "stardog.fullname" $ }}:5820 user passwd -N "$STARDOG_ADMIN_PW"
```
srueg: Please use the full argument names to make it more readable.

stardog/templates/post-install-add-users.yaml
janPhil: probably you mean port 80 instead of 5820 during the availability-check?

stardog/templates/post-install-add-users.yaml
janPhil: so instead of -N --new-password?

stardog/templates/post-install-add-users.yaml
srueg: Right now the curl command uses port 80 (implicitly since no port is specified). Please change it to also use port 5820.
And to make the whole thing more DRY, please create an env var with the URL and use it everywhere:
```
env:
  - name: STARDOG_SERVER_URL
    value: http://{{ include "stardog.fullname" $ }}:5820
```

stardog/templates/post-install-add-users.yaml
srueg: yes

PR #66, opened 2019-07-26
[Signalilo] Update to 0.1.0

corvus-ch:

Update to Signalilo 0.1.0.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/signalilo/values.yaml
+++ b/signalilo/values.yaml
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-replicaCount: 2
+replicaCount: 1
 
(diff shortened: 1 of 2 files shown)


signalilo/values.yaml
```diff
@@ -2,12 +2,12 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-replicaCount: 2
+replicaCount: 1
```
mhutter: If it's not intended to be scaled, why not remove the option alltogether?

mhutter: If it's not intended to be scaled, why not remove the option alltogether?

PR #67, opened 2019-08-15
[Keycloak] Deprecate chart

srueg:
The upstream chart from Codecentric now includes the OpenShift route.
Therefore there's no need to have our own chart anymore.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/keycloak/Chart.yaml
+++ b/keycloak/Chart.yaml
@@ -1,5 +1,6 @@
 name: keycloak
-version: 4.12.0
+version: 5.0.0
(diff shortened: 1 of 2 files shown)


keycloak/Chart.yaml
```diff
@@ -1,5 +1,6 @@
 name: keycloak
-version: 4.12.0
```
anothertobi: Deprecating this chart is kind of a breaking change. I'd prefer to update the major version.

PR #86, opened 2019-10-07
[openshift-route] Add openshift-route chart

ccremer:
#### Checklist

- [x] Chart Version bumped


openshift-route/README.md
```diff
@@ -0,0 +1,38 @@
...
+| Parameter | Description | Default
+| ---       | ---         | ---
+| `host` | The hostname that should be used. If left empty, OpenShift will generate one for you with defaults. | `""`
+| `nameOverride` | Overrides the name after the Chart name. Affects route name and hostname generation. | `""`
+| `fullnameOverride` | If you need to give a specific name to the route, override this, as it affects the hostname generated if `host` is left empty. | `""`
+| `path` | Subpath of the route. | `""`
+| `annotations` | Annotations on the route object. | `{}`
+| `service.name` | The backend service name of the route. Required. | `""`
+| `service.targetPort` | The port of the backend service. | `8080`
+| `service.weight` | Weight of the service endpoint. | `1`
```
srueg: I guess the default should be `100`?

openshift-route/values.yaml
```diff
@@ -0,0 +1,32 @@
...
+  insecureEdgeTerminationPolicy: Redirect
+  key:
+  certificate:
+  caCertificate:
+  destinationCACertificate:
+
+service:
+  name: "adf"
+  targetPort: 8080
+  weight: 1
```
srueg: Should be `100`

openshift-route/README.md
ccremer: Not according to documentation: https://docs.openshift.com/container-platform/3.9/architecture/networking/routes.html#alternateBackends, or at least as I interpret it.

srueg: Ah you're right. But since the OpenShift default if you don't set a weight is 100, I'd propose to also use this as default

PR #99, opened 2019-12-11
[stardog] Add user, role and database support

ccremer:
#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/stardog/files/stardog-wrapper.sh
+++ b/stardog/files/stardog-wrapper.sh
@@ -1,0 +1,118 @@
+#!/bin/bash
+
+# saner programming env: these switches turn some bugs into errors
+set -o errexit -o pipefail -o noclobber -o nounset
+
+# -allow a command to fail with !’s side effect on errexit
+# -use return value from ${PIPESTATUS[0]}, because ! hosed $?
+! getopt --test > /dev/null
+if [[ ${PIPESTATUS[0]} -ne 4 ]]; then
+    >&2 echo 'I’m sorry, `getopt --test` failed in this environment.'
+    exit 1
+fi
+
+OPTIONS=sdahn:p:v
+LONGOPTS=server,disable-anonymous,set-admin-password,help,new-user-name:,new-user-pass-from:,verbose
+
+# -regarding ! and PIPESTATUS see above
+# -temporarily store output to be able to check for errors
+# -activate quoting/enhanced mode (e.g. by writing out “--options”)
+# -pass arguments only via   -- "$@"   to separate them correctly
+! PARSED=$(getopt --options=$OPTIONS --longoptions=$LONGOPTS --name "$0" -- "$@")
+if [[ ${PIPESTATUS[0]} -ne 0 ]]; then
+    # e.g. return value is 1
+    #  then getopt has complained about wrong arguments to stdout
+    exit 2
+fi
+# read getopt’s output this way to handle the quoting right:
+eval set -- "$PARSED"
+
+function usage {
+	cat <<EOM
+Usage: $(basename "$0") [OPTION]...
+
+  -s,--server             VALUE  Stardog server URL. Uses \$STARDOG_SERVER_URL if unset.
+  -d,--disable-anonymous         Disable the anonymous user.
+  -a,--set-admin-password        Sets the admin password. Needs \$STARDOG_ADMIN_PW set as ENV var.
+  -v,--verbose                   Verbose mode.
+  -n,--new-user-name      VALUE  Adds a new user name with the given name. Needs -p.
+  -p,--new-user-pass-from PATH   When using --new-user-name, set a path to a file which contains the new user password.
+  -h,--help                      Display this help.
+EOM
+	exit 2
+}
+
+admin_bin="${STARDOG_ADMIN_BIN:-/opt/stardog/bin/stardog-admin}"
+admin_pw="${STARDOG_ADMIN_PW}"
+server="${STARDOG_SERVER_URL:-http://localhost:5820}"
+
+disable_anonymous=false
+set_admin_password=false
+verbose=false
+new_user=false
+
+# now enjoy the options in order and nicely split until we see --
+while true; do
+    case "$1" in
+        -s|--server)
+            server="${2}"
+            shift 2
+            ;;
+        -d|--disable-anonymous)
+            disable_anonymous=true
+            shift
+            ;;
+        -a|--set-admin-password)
+            set_admin_password=true
+            shift
+            ;;
+        -h|--help)
+            usage
+            exit 1
+            ;;
+        -n|--new-user-name)
+            new_user=true
+            new_user_name="${2}"
+            shift 2
+            ;;
+        -p|--new-user-pass-from)
+            new_user_pass_from="${2}"
+            shift 2
+            ;;
+        -v|--verbose)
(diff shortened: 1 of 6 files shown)


stardog/files/stardog-wrapper.sh
```diff
@@ -0,0 +1,164 @@
...
+     --new-database        VALUE                Creates a new database with the given name. Pass additional options after --,
+                                                e.g. "--new-database stardog -- -o search.enabled=true"
+                                                Will skip existing databases.
+  -h,--help                                     Display this help and exit.
+  -v,--verbose                                  Verbose mode.
+EOM
+	exit 2
+}
+
+admin_bin="${STARDOG_ADMIN_BIN:-/opt/stardog/bin/stardog-admin}"
```
srueg: A bash function could be used to always use `--passwd`:
```
stardog-admin () {
  $admin_bin $@ --passwd ${admin_pw}
}
```

Not sure where exactly the argument goes...

stardog/templates/post-install-job.yaml
```diff
@@ -35,41 +35,58 @@ spec:
...
-          {{ range .Values.stardog.users }}
-          {{- $pwEnv := .name | upper | replace "-" "_"}}
-          echo Adding user {{ .name }} to stardog
-          /opt/stardog/bin/stardog-admin --server "${STARDOG_SERVER_URL}" user add --new-password {{ print "\"${" $pwEnv "}\"" }} --passwd "${STARDOG_ADMIN_PW}" {{ .name }} || \
-          echo User already exists - setting password && \
-          /opt/stardog/bin/stardog-admin --server "${STARDOG_SERVER_URL}" user passwd --new-password {{ print "\"${" $pwEnv "}\"" }} --passwd "${STARDOG_ADMIN_PW}" {{ .name }}
-          echo
+          {{ if .Release.IsInstall -}}
+          bash ${script} -v --set-admin-password
+          {{- range .Values.stardog.disabledUsers }}
```
srueg: This should probably be outside the `if .Release.IsInstall`

stardog/values.yaml
```diff
@@ -23,6 +23,20 @@ stardog:
...
   # - name: nameOfTheUser
   #   password: passwordForTheUser
+  disabledUsers:
+  - anonymous
+  databases:
+  # - name: stardog
+  #   options:
+  #   - name: option
+  #     value: value
+  roles:
```
srueg: We could create a default `superuser` role to be used: https://www.stardog.com/docs/#_superusers

stardog/templates/post-install-job.yaml
```diff
@@ -35,41 +35,58 @@ spec:
...
+          until $(curl -o /dev/null -s -f "${STARDOG_SERVER_URL}/admin/alive"); do sleep 1; done
+          echo "Stardog is available now"
 
-          {{- if .Release.IsInstall }}
-          echo Disable anonymous user
-          /opt/stardog/bin/stardog-admin --server "${STARDOG_SERVER_URL}" user disable anonymous
-          echo Changing admin password
-          /opt/stardog/bin/stardog-admin --server "${STARDOG_SERVER_URL}" user passwd --new-password "$STARDOG_ADMIN_PW"
-          echo
-          {{- end }}
```
ccremer: @srueg In the previous version both were in IsInstall block, should that be changed?

srueg: If the call doesn't fail if disabling an already disabled user, yes, I'd propose to run it always.

stardog/files/stardog-wrapper.sh
ccremer: Hm, this would break scenarios where I need arguments at the end of the command to provide additional args for the command itself, e.g. `bash ${script} -v --new-database my-database -- -o "search.enabled=true"`

srueg: true, never mind then

stardog/values.yaml
ccremer: Hm, a superuser role is already present internally. This can be achieved easily with the superuser flag at user creation, e.g. `bash ${script} -v --ensure-user {{ .name }} --user-pass-from "/run/secrets/stardog/{{ .name }}" -- --superuser` would result in `${admin_bin} user add --new-password $(cat "${user_pass_from}") --passwd "${admin_pw}" "${user_name}" --superuser`. Should we can add the flag in the `.users` dict or create/provide a new role out of it?

srueg: Let's try it with the role, this way the superuser is just a regular role (and could be changed after user creation)

stardog/templates/post-install-job.yaml
ccremer: Need to check that first

ccremer: So, it works repeatedly with exit code 0. But then my question is, why not run the command to set the admin password every time as well? Basically remove the `if IsInstall` condition?

ccremer:
Tested with all features enabled again. this is the output of the Postinstall job for the first time installation:
```sh
helm upgrade --install stardog-dev  --namespace stardog-dev --set stardog.adminPassword=stardog --set image.username=xxx --set image.password=xxx --set persistence.enabled=false --set replicaCount=1 --set zookeeper.enabled=false stardog
```

```console
Testing stardog connectivity to http://stardog-dev:5820...
Stardog is available now
Setting admin password
Password changed successfully for user admin.
Disable user anonymous
User anonymous has been successfully disabled.
Checking if database stardog exists...
stardog does not exist
Successfully created database 'stardog'.

Adding user nameOfTheUser to stardog
Successfully created user nameOfTheUser.
Password changed successfully for user nameOfTheUser.
Creating role superuser if not existing
Successfully added role superuser.
Granting 'all' permission to role 'superuser' for '*:*'
Successfully granted the permission.
Assigning role superuser to user admin
Role superuser has been successfully assigned to user admin.
```
After upgrading the helm chart with new value (add nameOfTheUser to superuser):
```console
Testing stardog connectivity to http://stardog-dev:5820...
Stardog is available now
Disable user anonymous
User anonymous has been successfully disabled.
Checking if database stardog exists...
Database             : stardog
Status               : Online
Size                 : 0 triples
Queries              : None running
Open Connections     : 0
Open Transactions    : 0
Query Avg. Time      : 0.00 s
Query Rate           : 0.00 queries/sec
Plans Cached         : 0
Plan Cache Hit Ratio : 0.00%
Adding user nameOfTheUser to stardog
User already exists: nameOfTheUser.
User already exists - updating password
Password changed successfully for user nameOfTheUser.
Creating role superuser if not existing
Role already exists: superuser.
Granting 'all' permission to role 'superuser' for '*:*'
Successfully granted the permission.
Assigning role superuser to user admin
user admin already has role superuser
Assigning role superuser to user nameOfTheUser
Role superuser has been successfully assigned to user nameOfTheUser.
```
I think it works now. WDYT?

stardog/values.yaml
```diff
@@ -23,6 +23,20 @@ stardog:
...
+  #   options:
+  #   - name: search.enabled
+  #     value: "true"
+  roles:
+  - name: superuser
+    grants:
+    - action: all
+      object: "*:*"
+    users:
+    - admin
```
srueg: Let's remove this user. The Stardog `admin` user is a special case which has `superuser` rights anyway.

PR #110, opened 2020-01-31
[Trifid] High Available

srueg:

* Upgrade Trifid to latest
* Make trifid highly available

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/trifid/templates/deployment.yaml
+++ b/trifid/templates/deployment.yaml
@@ -60,10 +60,20 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-    {{- with .Values.affinity }}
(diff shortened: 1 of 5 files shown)


trifid/templates/deployment.yaml
```diff
@@ -60,10 +60,20 @@ spec:
       nodeSelector:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-    {{- with .Values.affinity }}
       affinity:
-        {{- toYaml . | nindent 8 }}
-    {{- end }}
+        podAntiAffinity:
```
ccremer: Why is this hardcoded in the template now? Why not make it default in the values.yaml? This way it's not customizable at all.

PR #131, opened 2020-02-20
Ignore test dirs in packaged helm releases

ccremer:

We often get PRs from Renovate for Go modules which fail in pipeline since the chart version was not bumped. I do not see a reason why the chart should be shipped with the Go testing files, by ignoring the test in the CI build, the build won't fail in the future, so we can safely merge PRs from Renovate without a Chart version bump.

See e.g. #130 

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -71,7 +71,7 @@ for chart in $charts; do
   chart=${chart:2}
   chart_changed=false
   for file in $changed_files; do
-    if [[ $file == "$chart/"* ]]; then
+    if [[ $file != "$chart/test/"* && $file == "$chart/"* ]]; then
       chart_changed=true
       break
     fi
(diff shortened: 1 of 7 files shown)


srueg:
Unfortunately that won't be enough to make the build work...
The `.ci/build.sh` script implements some logic to see if the files of a chart changed and will fail the build if the version of the chart wasn't bumped.
While I like this change, the build.sh script needs changing as well.

ccremer:
I managed to get it working (https://travis-ci.com/appuio/charts/builds/149788778). Now it ignores $charts/test/* files, but adding the `test/` entry in the `.helmignore` still triggers a change in the chart, so lets add the helmignore entry, bump the chart, merge this PR and see if #130 works again.

PR #132, opened 2020-02-20
Include helm test in CI pipelines

ccremer:

* Updates Helm repositories based on requirements.yaml
* Runs go test if there is a test dir

#### Checklist

- [ ] Chart Version bumped

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -55,7 +55,8 @@ cleanup() {
 }
 trap cleanup 0 1 2 3 6 15
 
-# Dependency repos
+# Dependency repos (we still have to download them manually, even if they are added in requirements.yaml)
+# (https://github.com/helm/helm/issues/6005)
 helm repo add bitnami https://charts.bitnami.com
 helm repo add codecentric https://codecentric.github.io/helm-charts
 helm repo add jetstack https://charts.jetstack.io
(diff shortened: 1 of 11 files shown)


.ci/build.sh
```diff
@@ -85,6 +92,7 @@ for chart in $charts; do
   any_chart_changed=true
 
   echo "----> Build dependencies for ${chart}"
+  helm dependency update
```
srueg: I don't think this will work as expected. The `dependency update` command it used to update the `requirements.lock` file and to actually update dependencies. So this shouldn't run in the pipeline. Running `dependency build` should be enough.

srueg: There is an open issue regarding this: https://github.com/helm/helm/issues/7214

.travis.yml
```diff
@@ -1,12 +1,15 @@
 ---
-language: minimal
+language: go
+go:
+  - "1.12"
```
srueg: Let's try with go 1.13

.ci/build.sh
```diff
@@ -70,6 +71,15 @@ echo "Changed files: $changed_files"
 for chart in $charts; do
   chart=${chart:2}
   chart_changed=false
+
+  # Run go test if there are chart unit tests
+  if [[ -f "$chart/test/go.mod" ]]; then
+    pushd "$chart/test"
+    echo "----> Running unit tests for ${chart}"
+    go mod download
```
srueg: explicitly downloading the modules is not necessary, it will be done automatically with `go test`

ccremer:
@srueg It's working now. Wanna review?

PR #138, opened 2020-03-05
[stardog] Enable the config option spatial.enabled

soufianebenali:
as requested we are enabling the option spatial.enabled for all instances of the cluster.

Resolves: SBAR-191

#### Checklist

- [x] Chart Version bumped


srueg:
As discussed, this is not an option that can be enabled globally, but needs to be set for each database.

PR #146, opened 2020-03-14
data-cube-curation chart

sandhose:
I wrote a chart for https://github.com/zazuko/data-cube-curation (see ticket SBAR-178)

It's still kinda WIP (mostly the readme, some missing variables and the registry not being the right one), but I'm waiting on some changes on data-cube-curation to have the UI working correctly.

I chose to split the secrets into multiple ones (with the ability to use an existing one for each one of them), not sure if it's the right way to do it

### DIFF
--- a/data-cube-curation/Chart.yaml
+++ b/data-cube-curation/Chart.yaml
@@ -1,0 +1,9 @@
+apiVersion: v2
+name: data-cube-curation
+description: RDF Data Cube curation service
+type: application
+
+version: 0.1.0
+# Right now the docker tags are git hashes
+appVersion: a74c6f116619f7cb2cd58bfd2b5f5db03b3df66b
+home: https://github.com/zazuko/data-cube-curation
--- a/data-cube-curation/Chart.yaml
+++ b/data-cube-curation/Chart.yaml
@@ -3,5 +3,5 @@ name: data-cube-curation
 description: RDF Data Cube curation service
 
 version: 0.1.0
-appVersion: latest
+appVersion: 0.2.0
 home: https://github.com/zazuko/data-cube-curation
--- a/data-cube-curation/Chart.yaml
+++ b/data-cube-curation/Chart.yaml
@@ -1,9 +1,7 @@
-apiVersion: v2
+apiVersion: v1
 name: data-cube-curation
 description: RDF Data Cube curation service
-type: application
 
 version: 0.1.0
-# Right now the docker tags are git hashes
-appVersion: a74c6f116619f7cb2cd58bfd2b5f5db03b3df66b
+appVersion: latest
 home: https://github.com/zazuko/data-cube-curation
--- a/data-cube-curation/templates/deployment.yaml
+++ b/data-cube-curation/templates/deployment.yaml
@@ -1,0 +1,152 @@
+{{- $fullName := include "data-cube-curation.fullname" . -}}
+{{- $sparqlSecretName := include "data-cube-curation.sparql.secretName" . -}}
+{{- $gitlabSecretName := include "data-cube-curation.gitlab.secretName" . -}}
+{{- $s3SecretName := include "data-cube-curation.s3.secretName" . -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ $fullName }}
+  labels:
+    {{- include "data-cube-curation.labels" . | nindent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "data-cube-curation.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "data-cube-curation.selectorLabels" . | nindent 8 }}
+    spec:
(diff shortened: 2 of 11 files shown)


data-cube-curation/Chart.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v2
```
srueg: Let's stick with `apiVersion: v1` since our tooling is not yet Helm v3 ready

data-cube-curation/templates/deployment.yaml
```diff
@@ -0,0 +1,152 @@
...
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: http
+              containerPort: 3000
+              protocol: TCP
+          livenessProbe:
```
srueg: An `initialDelaySeconds` probably makes sense here, to give the pod some time to start up

sandhose:
@srueg Did the changes you requested and tweaked the chart a bit (see last commit)

data-cube-curation/Chart.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+name: data-cube-curation
+description: RDF Data Cube curation service
+
+version: 0.1.0
+appVersion: latest
```
srueg: Let's use a proper version number here. Otherwise it will get hard to track versions and do rollbacks etc.

data-cube-curation/templates/deployment.yaml
```diff
@@ -0,0 +1,172 @@
...
+            httpGet:
+              path: /
+              port: http
+            initialDelaySeconds: 30
+          readinessProbe:
+            httpGet:
+              path: /
+              port: http
+            initialDelaySeconds: 30
```
srueg: This should be shorter (e.g. 3s) I suggest, otherwise it will take at least 30s for new pods to become ready (receive traffic).

data-cube-curation/values.yaml
```diff
@@ -0,0 +1,84 @@
...
+    # kubernetes.io/tls-acme: "true"
+  hosts:
+    - host: data-cube-curation.localhost
+      paths: [/]
+  tls: []
+  #  - secretName: chart-example-tls
+  #    hosts:
+  #      - chart-example.local
+
+resources: {}
```
srueg: Is there an expected baseline expected for this app? If so please configure it as the default.

data-cube-curation/Chart.yaml
sandhose: It seems like there are no stable tags/version number right now. Should I use the sha256 digest of the container for now? Or tag an arbitrary v0.0.1?

data-cube-curation/values.yaml
sandhose: I did that because the [default template](https://github.com/helm/helm/blob/fd00ca260a7132bc2f4678f1ff2f8253feac317c/pkg/chartutil/create.go#L140-L150) (when running `helm create [chart]`) has this comment on it:

```yaml
resources: {}
  # We usually recommend not to specify default resources and to leave this as a conscious
  # choice for the user. This also increases chances charts run on environments with little
  # resources, such as Minikube. If you do want to specify resources, uncomment the following
  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
  # limits:
  #   cpu: 100m
  #   memory: 128Mi
  # requests:
  #   cpu: 100m
  #   memory: 128Mi
```

data-cube-curation/templates/deployment.yaml
sandhose: Considering that the `periodSeconds` defaults to 10 and the `failureThreshold` to 3 (see [reference](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#probe-v1-core)), the container can take up to 30s before getting killed without the initial delay. Not sure if adding 3s to that makes much sense.

data-cube-curation/Chart.yaml
srueg: If you can I'd tag the v0.0.1, yes. Otherwise let's go with the digest.

data-cube-curation/templates/deployment.yaml
srueg: Yes agree, for the readiness probe we can get rid of the delay

data-cube-curation/values.yaml
srueg: Ok, fine by me. I usually try to come up with a sensible baseline for applications I'm able to estimate.

data-cube-curation/templates/deployment.yaml
sandhose: Done

data-cube-curation/Chart.yaml
sandhose: Done

sandhose:
Should be ready to merge now

PR #151, opened 2020-03-23
[k8up] Modernize chart

ccremer:

- Adopt Helm conventions (breaking change)
- Make indents consistent

#### Checklist

- [x] Chart Version bumped


k8up/templates/deployment.yaml
```diff
@@ -1,36 +1,63 @@
...
-    app: {{ template "k8up-operator.name" $ }}
-    chart: {{ template "k8up-operator.chart" $ }}
-    release: {{ .Release.Name }}
-    heritage: {{ .Release.Service }}
+{{ include "k8up.labels" . | indent 4 }}
 spec:
+  # Running more than 1 replica is not supported, thus not configurable.
   replicas: 1
+  strategy:
+    type: recreate
```
srueg:
```suggestion
    type: Recreate
```

k8up/templates/deployment.yaml
```diff
@@ -1,36 +1,63 @@
...
+          {{- end }}
           ports:
-          - containerPort: 8080
-            protocol: TCP
+            - name: http
+              containerPort: 8080
+          livenessProbe:
+            httpGet:
+              path: /metrics
+              port: http
```
srueg: Let's add an `initialDelaySeconds`

k8up/templates/deployment.yaml
```diff
@@ -1,36 +1,63 @@
...
+      {{- with .Values.affinity }}
+      affinity: {{ . }}
+      {{- end }}
+      serviceAccountName: {{ template "k8up.serviceAccountName" . }}
+      {{- with .Values.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    {{- with .Values.affinity }}
+      affinity:
```
srueg: Duplicate?

k8up/values.yaml
```diff
@@ -1,7 +1,21 @@
...
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name:
+
+k8up:
   ## Allows the specification of additional environment variables
+  extraEnvVars: []
```
srueg: What's the point of the `extraEnvVars` ?

k8up/templates/deployment.yaml
ccremer: will be fixed in next commit

k8up/templates/deployment.yaml
ccremer: yes, noticed that too

k8up/values.yaml
ccremer: I dunno. Was there before. The only reason I can imagine is that we can leave extraEnv with `BACKUP_IMAGE` variable unmodified if one wants to add more variables. But if `BACKUP_IMAGE` is required anyway, we might as well include that fixed in the deployment and provide an own helm variable just for the backup image. WDYT @Kidswiss ?

Kidswiss: A lot of configuration can be passed to the operator via the envvars. But the only one that's mandatory is the image version, so it makes sense to add this as a helm variable.

ccremer:
Forgot the docs, one sec

ccremer:
I'm ready for a review now. Please indicate if the level of unit test is sufficient or if I should go deeper (e.g. did not create any tests for RBAC, as there is nothing to configure really)

k8up/templates/deployment.yaml
```diff
@@ -1,36 +1,63 @@
...
           ports:
-          - containerPort: 8080
-            protocol: TCP
+            - name: http
+              containerPort: 8080
+          livenessProbe:
+            httpGet:
+              path: /metrics
+              port: http
+            initialDelaySeconds: 2
```
srueg: I think that's a little bit too optimistic. Let's put at least 30s so if the operator takes a while starting up it's not ending up in a kill loop.

srueg:
Very nice! Just a small nitpick.

k8up/templates/deployment.yaml
ccremer: I have no idea how long the startup takes. I was under impression that the prometheus endpoint should be up fairly quickly. Yes there might be stuff in the background that still needs warmup. @Kidswiss how long does a startup usually take? Are 30s justified?

srueg: True but regardless, I don't hink a delay shorter than 30s ever makes sense. Think about a cluster which is under some resource pressure and it just takes a while to start serving traffic. In such a situation it makes matters worse if the pod keeps getting killed...

PR #169, opened 2020-05-07
[k8up] Add Prometheus monitoring manifests and rules

ccremer:

* Adds prometheus Rules and manifests for the Prometheus Operator
* Internal Reference: APPU-2567

#### Checklist

- [x] Chart Version bumped


k8up/templates/prometheus/prometheusrule.yaml
```diff
@@ -0,0 +1,41 @@
...
+kind: PrometheusRule
+metadata:
+  name: {{ include "k8up.fullname" . }}-rule
+  labels:
+    {{- include "k8up.labels" . | nindent 4 }}
+spec:
+  groups:
+    - name: K8up
+      rules:
+      - alert: baas_last_errors
```
srueg, line 12: Let's keep naming consistent
```suggestion
      - alert: K8upResticErrors
```

ccremer:
I wonder whether we should keep the default rules in the PrometheusRule ("hardcoded") and enable a key like `metrics.prometheusRule.additionalRules`, Or move those as defaults to the values.yaml file in the key `metrics.prometheusRule.rules`. WDYT?

k8up/templates/prometheus/prometheusrule.yaml
```diff
@@ -0,0 +1,41 @@
+{{- if .Values.metrics.prometheusRule.enabled -}}
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: {{ include "k8up.fullname" . }}-rule
```
srueg, line 5:
```suggestion
  name: {{ include "k8up.fullname" . }}-rule
  namespace: {{ default .Values.metrics.prometheusRule.namespace .Release.Namespace }}
```

srueg:
Good point. I'd suggest leaving the default rules in the file but with an `if` around them. And to provide a `metrics.prometheusRule.additionalRules` value to add custom rules.

ccremer:
I'm adding some unit tests soon but the change is now ready for discussions

ccremer:
Hi @Kidswiss 
I finally managed to get prom operator and K8up running on local K3s. However, the metrics on the K8up operator pod does not show up. I'm missing specifically those for the alert rules:
```
- k8up_jobs_failed_counter
- baas_backup_restic_last_errors
- k8up_jobs_total
- k8up_jobs_queued_gauge
- k8up_schedules_gauge
```
Do these only show up when there were backups?

ccremer:
@srueg I noticed that the ServiceMonitor object only is picked up by prometheus if there's the metrics service enabled, for that all
* `metrics.enabled=true`
* `metrics.service.enabled=true` and 
* `metrics.serviceMonitor.enabled=true` 

are necessary. Without the Service, the ServiceMonitor does nothing.

How should we solve this? I see several approaches:
1. Implicitly render the Service template too when enabling `metrics.serviceMonitor.enable=true` (boolean OR with (`metrics.enabled=true` and `metrics.service.enabled=true`)
2. Require that `metrics.service.enabled` needs also to be true in order for ServiceMonitor to be rendered (boolean AND with `metrics.serviceMonitor.enabled=true`)
3. Use a PodMonitor instead of service, though that probably requires a recent prom operator to be installed, right?

Grepping through the `metrics.enabled` flag, I see that the only occurrence of this flag in the templates is in the service template, ANDed with `metrics.service.enabled`, which feels redundant. I think we can actually get rid of either `metrics.service.enabled` or `metrics.enabled`

srueg:
Agree, let's get rid of the `metrics.service.enabled` value.

> Use a PodMonitor instead of service, though that probably requires a recent prom operator to be installed, right?

Exactly, unfortunately that's not available on OpenShift nor Rancher.

ccremer:
> Agree, let's get rid of the `metrics.service.enabled` value.

Do you mean to always deploy the service object or keep `metrics.enabled` on just for this service, even though we have separate switches for prometheusRule and serviceMonitor?

srueg:
Let's always deploy the service.

ccremer:
Thanks for the review. This is a MWE with Helmfile to get it working on local K3s, for future reference:
```yaml
repositories:
- name: stable
  url: https://kubernetes-charts.storage.googleapis.com
- name: appuio
  url: https://charts.appuio.ch

releases:
- name: appuio-k8up
  namespace: k8up
  chart: appuio/k8up
  values:
  - metrics:
      serviceMonitor:
        enabled: true
      prometheusRule:
        enabled: true
- name: appuio-prometheus-operator
  namespace: prometheus
  chart: stable/prometheus-operator
  values:
  - prometheus:
      prometheusSpec:
        serviceMonitorSelectorNilUsesHelmValues: false
        ruleSelectorNilUsesHelmValues: false
```
However it may make sense to install the chart with `helm install -n prometheus appuio-prometheus-operator stable/prometheus-operator` first in order to get the CRDs installed, otherwise `helmfile diff` fails to render a diff because of missing CRDs.

Kidswiss:
> Hi @Kidswiss
> I finally managed to get prom operator and K8up running on local K3s. However, the metrics on the K8up operator pod does not show up. I'm missing specifically those for the alert rules:
> 
> ```
> - k8up_jobs_failed_counter
> - baas_backup_restic_last_errors
> - k8up_jobs_total
> - k8up_jobs_queued_gauge
> - k8up_schedules_gauge
> ```
> 
> Do these only show up when there were backups?

Yes the metric will only start to show up once there's a backup. `baas_backup_restic_last_errors` will be pushed to a prometheus push gateway that needs to be scraped by the prometheus instance.

Kidswiss:
The rules themselves LGTM.

PR #179, opened 2020-06-08
[data-cube-curation] use chart appversion as image tag & update the app

sandhose:

This updates the data-cube-curation app to a newer version. It also changes the `image.tag` value to use the `Chart.AppVersion` by default.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/data-cube-curation/values.yaml
+++ b/data-cube-curation/values.yaml
@@ -34,7 +34,7 @@ replicaCount: 1
 
 image:
   repository: docker.io/zazuko/datacube
-  tag: api_0.2.0
+  tag: "{{ .Chart.AppVersion }}"
   pullPolicy: IfNotPresent
 
 imagePullSecrets: []
--- a/data-cube-curation/Chart.yaml
+++ b/data-cube-curation/Chart.yaml
@@ -2,6 +2,6 @@ apiVersion: v1
 name: data-cube-curation
 description: RDF Data Cube curation service
 
-version: 0.1.1
-appVersion: 0.2.0
+version: 0.1.2
+appVersion: 0.3.0
 home: https://github.com/zazuko/data-cube-curation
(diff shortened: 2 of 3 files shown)


data-cube-curation/values.yaml
```diff
@@ -34,7 +34,7 @@ replicaCount: 1
 
 image:
   repository: docker.io/zazuko/datacube
-  tag: api_0.2.0
+  tag: "{{ .Chart.AppVersion }}"
```
ccremer, line 37: The one downside of this change would be that Renovate (a bot that automatically creates new PRs if there are newer docker tags available) will stop working for this chart, thus stop notifying us about new releases, because I don't think it can parse the `{{ .Chart.AppVersion }}` expression as a semver number.

Personally I'm against that change and accept a little duplication in that case. But I don't have a hard feeling either for this chart :)

ccremer: WDYT @srueg ?

ccremer: Relevant documentation: https://docs.renovatebot.com/modules/manager/helm-values/

srueg: Hmm good point. I'd also recommend letting Renovate do it's job.
Since this is mostly managed by Zazuko it's up to you @sandhose

sandhose: Well, if we let renovate update the image tag, it means the appVersion field will be out of sync (as renovate does not seem to update it) ; not that it's very important, but I think I prefere updating the appVersion manually every once in a while

ccremer: I understand you point. However, when Renovate creates a PR with the new image tag, we have to touch the PR in any case, as Renovate does not update `version` in `chart.yaml`. A change in the helm chart without updating the chart version will fail the pipeline so we cannot merge it either way. Maybe that changes in the future when Renovate supports that, but for now it's "okay", at least we get notified automatically.

We can accept your change, no problem. But then you would have to create PRs on your own if the chart should be updated with newer image tags, as we don't check for new releases every day :)

sandhose: > We can accept your change, no problem. But then you would have to create PRs on your own if the chart should be updated with newer image tags, as we don't check for new releases every day :)

Sounds good to me 👍

PR #186, opened 2020-06-30
[Stardog] Only run postinstall if replicas > 0

chamer81:
This prevents the postinstall job from running when the chart is installed with replicaCount equal to zero. The postinstall job requires at least one running pod in order to terminate successfully, so setting the replica count to zero via gitops would always lead to a fail state without this change.
#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/stardog/test/postinstall_test.go
+++ b/stardog/test/postinstall_test.go
@@ -1,0 +1,42 @@
+package test
+
+import (
+	"testing"
+
+	batchv1 "k8s.io/api/batch/v1"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/stretchr/testify/assert"
+)
+
+var tplPostInstallJob = []string{"templates/post-install-job.yaml"}
+
+func Test_PostInstallJob_Should_Not_Exist(t *testing.T) {
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"stardog.adminPassword":     adminPassword,
+			"replicaCount":	"0",
+		},
+	}
+
+	output, err := helm.RenderTemplateE(t, options, helmChartPath, releaseName, tplPostInstallJob)
+
+	assert.NotEmpty(t, err)
+	assert.Equal(t, "Error: could not find template templates/post-install-job.yaml in chart", output)
+}
+
+func Test_PostInstallJob_Should_Exist(t *testing.T) {
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"stardog.adminPassword":     adminPassword,
+			"replicaCount":	"1",
+		},
+	}
+
+	output := helm.RenderTemplate(t, options, helmChartPath, releaseName, tplPostInstallJob)
+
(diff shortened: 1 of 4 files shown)


srueg:
Can you please add a test in the `tests/` folder to verify the behavior?
And I think you can target the `master` branch with this PR.

chamer81:
I have added the unit tests. But since I created this branch from the stardog/update-to-v7 branch, I can't merge it into master without bringing along all of those changes -- unless the idea is to cherry-pick the commits to a new branch from master. @srueg what do you suggest?

srueg:
Yes, you can reset this branch to `master` and then cherry-pick the two commits for this change 👍

stardog/test/postinstall_test.go
```diff
@@ -0,0 +1,42 @@
...
+	options := &helm.Options{
+		SetValues: map[string]string{
+			"stardog.adminPassword":     adminPassword,
+			"replicaCount":	"0",
+		},
+	}
+
+	output, err := helm.RenderTemplateE(t, options, helmChartPath, releaseName, tplPostInstallJob)
+
+	assert.NotEmpty(t, err)
```
ccremer, line 24: The `assert.NoError(t, err)` is better suited for testing errors. please use that.

stardog/test/postinstall_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	batchv1 "k8s.io/api/batch/v1"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/stretchr/testify/assert"
+)
+
+var tplPostInstallJob = []string{"templates/post-install-job.yaml"}
+
+func Test_PostInstallJob_Should_Not_Exist(t *testing.T) {
```
ccremer, line 14: A more fitting title: `Test_PostInstallJob_GivenReplicaCount_WhenZero_ThenDoNotRenderJob`

stardog/test/postinstall_test.go
```diff
@@ -0,0 +1,42 @@
...
+		},
+	}
+
+	output, err := helm.RenderTemplateE(t, options, helmChartPath, releaseName, tplPostInstallJob)
+
+	assert.NotEmpty(t, err)
+	assert.Equal(t, "Error: could not find template templates/post-install-job.yaml in chart", output)
+}
+
+func Test_PostInstallJob_Should_Exist(t *testing.T) {
```
ccremer, line 28: A more fitting title: `Test_PostInstallJob_GivenReplicaCount_WhenGreaterThanZero_ThenRenderJob`

stardog/test/postinstall_test.go
srueg: It's the exact opposite: `assert.Error(t, err)`

ccremer: riiight, mixed that up :)

chamer81:
Replaced by #187

PR #194, opened 2020-08-18
[k8up] Bump default wrestic version

Kidswiss:

Bumping default wrestic to a newer version.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/k8up/Chart.yaml
+++ b/k8up/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
   - backup
   - operator
   - appuio
-version: 0.6.1
+version: 0.6.2
 appVersion: v0.1.10
 sources:
   - https://github.com/vshn/k8up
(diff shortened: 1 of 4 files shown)


ccremer (changes requested):
The CI Build fails with 
```
ERROR] templates/clusterrole.yaml: the kind "rbac.authorization.k8s.io/v1beta1 ClusterRole" is deprecated in favor of "rbac.authorization.k8s.io/v1 ClusterRole"
[ERROR] templates/clusterrolebinding.yaml: the kind "rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding" is deprecated in favor of "rbac.authorization.k8s.io/v1 ClusterRoleBinding"
```
Can you change from `v1beta1` to `v1` in the affected files? I already verified that `rbac.authorization.k8s.io/v1` is available on OpenShift 3.11, so it should be compatible.

PR #200, opened 2020-10-02
[Stardog] Verify backups

chamer81:
The only way to tell if a given Stardog backup is valid is to try restoring it. This PR adds verification steps to the backup cronjob to do the following after performing a backup:

- Restore the backup with a temporary name
- Confirm that the restored database has the same number of triples as the original database
- Drop the restored database

If the backup can't be restored with the same number of triples, the cronjob fails.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/stardog/Chart.yaml
+++ b/stardog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: stardog
-version: 0.5.0
+version: 0.6.0
 appVersion: 7.3.2
 description: Stardog Helm Chart
 home: "https://www.stardog.com/"
--- a/stardog/templates/cronjob.yaml
+++ b/stardog/templates/cronjob.yaml
@@ -68,11 +68,21 @@ spec:
             - -c
             - |
               set -e
-              BACKUP_TARGET="s3://${S3_ENDPOINT}/${S3_BUCKET_NAME}/${S3_PATH_PREFIX}/$(date --utc +%Y-%m-%d_%H%M)?AWS_ACCESS_KEY_ID=${S3_ACCESS_KEY}&AWS_SECRET_ACCESS_KEY=${S3_SECRET_KEY}&region=${S3_REGION}"
-              /opt/stardog/bin/stardog-admin db status --passwd "${STARDOG_ADMIN_PW}" {{ $db.name | quote }}
-              /opt/stardog/bin/stardog-admin db backup --passwd "${STARDOG_ADMIN_PW}" \
-                --to "${BACKUP_TARGET}" \
-                {{ $db.name | quote }}
+              BACKUP_DATE=$(date --utc +%Y-%m-%d_%H%M)
+              DATABASE_NAME={{ $db.name | quote }}
+              CURRENT_TRIPLE_COUNT=$(/opt/stardog/bin/stardog data size --exact "${DATABASE_NAME}" --passwd "${STARDOG_ADMIN_PW}" | cut -d " " -f 2 | tr -d ",")
+              echo "${DATABASE_NAME} has ${CURRENT_TRIPLE_COUNT} triples."
+              BACKUP_TARGET="s3://${S3_ENDPOINT}/${S3_BUCKET_NAME}/${S3_PATH_PREFIX}/${BACKUP_DATE}?AWS_ACCESS_KEY_ID=${S3_ACCESS_KEY}&AWS_SECRET_ACCESS_KEY=${S3_SECRET_KEY}&region=${S3_REGION}"
+              /opt/stardog/bin/stardog-admin db status --passwd "${STARDOG_ADMIN_PW}" "${DATABASE_NAME}"
+              /opt/stardog/bin/stardog-admin db backup --passwd "${STARDOG_ADMIN_PW}" --to "${BACKUP_TARGET}" "${DATABASE_NAME}"
+              RESTORE_TARGET="s3://${S3_ENDPOINT}/${S3_BUCKET_NAME}/${S3_PATH_PREFIX}/${BACKUP_DATE}/${DATABASE_NAME}?AWS_ACCESS_KEY_ID=${S3_ACCESS_KEY}&AWS_SECRET_ACCESS_KEY=${S3_SECRET_KEY}&region=${S3_REGION}"
+              RESTORE_NAME="${DATABASE_NAME}_${BACKUP_DATE}"
+              /opt/stardog/bin/stardog-admin db restore --passwd "${STARDOG_ADMIN_PW}" --name "${RESTORE_NAME}" "${RESTORE_TARGET}"
+              RESTORED_TRIPLE_COUNT=$(/opt/stardog/bin/stardog data size --exact "${RESTORE_NAME}" --passwd "${STARDOG_ADMIN_PW}" | cut -d " " -f 2 | tr -d ",")
+              echo "After backup-and-restore: ${RESTORED_TRIPLE_COUNT} triples."
+              /opt/stardog/bin/stardog-admin db drop --passwd "${STARDOG_ADMIN_PW}" "${RESTORE_NAME}"
+              SAME_COUNT=$((${CURRENT_TRIPLE_COUNT} == ${RESTORED_TRIPLE_COUNT} ? 0 : 1 ))
+              exit ${SAME_COUNT}
           restartPolicy: Never
           imagePullSecrets:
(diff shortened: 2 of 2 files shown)


stardog/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: stardog
-version: 0.5.0
+version: 0.5.1
```
ccremer, line 3: Since this is a new feature and not a bugfix I'd change that to 0.6.0

ccremer (changes requested):

stardog/Chart.yaml
chamer81: Yes, good point. I have made the update.

ccremer:
Is there some way we know beforehand about the restored disk size? Do we need to check that first or can we use a temporary disk location (emptyDir, hostpath, etc). If not, that basically means we have to provision double disk space even if it's just for peak usage.

chamer81:
> Is there some way we know beforehand about the restored disk size? Do we need to check that first or can we use a temporary disk location (emptyDir, hostpath, etc). If not, that basically means we have to provision double disk space even if it's just for peak usage.

Yes, we can get a good idea of how much space will be required: 

The Stardog server saves the data of each database in an individual folder. The restored db should be the same size as the origins. So we can just check the size of the current folder and then make sure we have that much space left on the PV (plus a little extra).

If the procedure fails due to not having enough space on the PV, the backup cronjob will exit in an error state, so we would know to take care of it. But Stardog is not good about automatically cleaning up failed restores, so that type of fail would actually leave the Stardog server in a bad state (stoarage area almost full due to containing broken leftovers from the failed restore.

So it would probably be better to check the size first and fail before attempting the restore if there's not enough space.

ccremer:
> If the procedure fails due to not having enough space on the PV, the backup cronjob will exit in an error state, so we would know to take care of it. But Stardog is not good about automatically cleaning up failed restores, so that type of fail would actually leave the Stardog server in a bad state (stoarage PV almost full due to containing broken leftovers from the failed restore).
> 
> So it would probably be better to check the size first and fail before attempting the restore if there's not enough space.

Right, this brings another topic: How do we ensure enough disk space?
* After a failed backup/restore, Admins probably want to investigate why it failed and need the data. So we keep the data in case of failure
* But for the next backup, we should ensure that the temporary db or files have been cleaned, otherwise it will fail immediately again. Is it ok to cleanup any previous DB restores *before* we try to backup/restore and only clean up afterwards on successful restores?

chamer81:
>     * After a failed backup/restore, Admins probably want to investigate why it failed and need the data. So we keep the data in case of failure
I'm not sure it would be useful to keep the data of the failed restore. The failed restore is simply a signal that really the backup silently failed. The bad backup data is stored on AWS -- that's the data the admins will want to investigate.

>     * But for the next backup, we should ensure that the temporary db or files have been cleaned, otherwise it will fail immediately again. Is it ok to cleanup any previous DB restores _before_ we try to backup/restore and only clean up afterwards on successful restores? I'm thinking of temporary failures here that might be successful on the next run (network problems etc.)

If we start getting a lot of false positives (i.e. the backup is actually fine, but the cronjob fails to restore the data correctly), then it would make sense to do more optimisation to eliminate the false positives. But my experience has been that the restore of a vaild backup will succeed as long as there's enough space and the configuration (credentials, etc.) is correct.

I lean towards adding a simple preliminary memory check on this iteration, and defer further optimisation for when it becomes necessary.

ccremer:
Yes, that makes more sense.
So we should make sure the PV is left in a clean state regardless of the outcome.
If this script is is becoming more complex, maybe it makes sense to have it mounted as a script via ConfigMap, similar to Stardog-wrapper.sh

stardog/templates/cronjob.yaml
```diff
@@ -68,11 +68,21 @@ spec:
...
-                {{ $db.name | quote }}
+              BACKUP_DATE=$(date --utc +%Y-%m-%d_%H%M)
+              DATABASE_NAME={{ $db.name | quote }}
+              CURRENT_TRIPLE_COUNT=$(/opt/stardog/bin/stardog data size --exact "${DATABASE_NAME}" --passwd "${STARDOG_ADMIN_PW}" | cut -d " " -f 2 | tr -d ",")
+              echo "${DATABASE_NAME} has ${CURRENT_TRIPLE_COUNT} triples."
+              BACKUP_TARGET="s3://${S3_ENDPOINT}/${S3_BUCKET_NAME}/${S3_PATH_PREFIX}/${BACKUP_DATE}?AWS_ACCESS_KEY_ID=${S3_ACCESS_KEY}&AWS_SECRET_ACCESS_KEY=${S3_SECRET_KEY}&region=${S3_REGION}"
+              /opt/stardog/bin/stardog-admin db status --passwd "${STARDOG_ADMIN_PW}" "${DATABASE_NAME}"
+              /opt/stardog/bin/stardog-admin db backup --passwd "${STARDOG_ADMIN_PW}" --to "${BACKUP_TARGET}" "${DATABASE_NAME}"
+              RESTORE_TARGET="s3://${S3_ENDPOINT}/${S3_BUCKET_NAME}/${S3_PATH_PREFIX}/${BACKUP_DATE}/${DATABASE_NAME}?AWS_ACCESS_KEY_ID=${S3_ACCESS_KEY}&AWS_SECRET_ACCESS_KEY=${S3_SECRET_KEY}&region=${S3_REGION}"
+              RESTORE_NAME="${DATABASE_NAME}_${BACKUP_DATE}"
```
srueg, line 79: Let's add a name to make it clear what this DB is about.

```suggestion
              RESTORE_NAME="${DATABASE_NAME}_restore_test_${BACKUP_DATE}"
```

srueg (changes requested):

srueg (approved):

stardog/templates/cronjob.yaml
```diff
@@ -68,16 +68,35 @@ spec:
...
-                --to "${BACKUP_TARGET}" \
-                {{ $db.name | quote }}
+              BACKUP_DATE=$(date --utc +%Y-%m-%d_%H%M)
+              DATABASE_NAME={{ $db.name | quote }}
+              CURRENT_TRIPLE_COUNT=$(/opt/stardog/bin/stardog data size --exact "${DATABASE_NAME}" --passwd "${STARDOG_ADMIN_PW}" | cut -d " " -f 2 | tr -d ",")
+              echo "${DATABASE_NAME} has ${CURRENT_TRIPLE_COUNT} triples."
+              BACKUP_TARGET="s3://${S3_ENDPOINT}/${S3_BUCKET_NAME}/${S3_PATH_PREFIX}/${BACKUP_DATE}?AWS_ACCESS_KEY_ID=${S3_ACCESS_KEY}&AWS_SECRET_ACCESS_KEY=${S3_SECRET_KEY}&region=${S3_REGION}"
+              /opt/stardog/bin/stardog-admin db status --passwd "${STARDOG_ADMIN_PW}" "${DATABASE_NAME}"
+              /opt/stardog/bin/stardog-admin db backup --passwd "${STARDOG_ADMIN_PW}" --to "${BACKUP_TARGET}" "${DATABASE_NAME}"
+              PERCENT_USE=$(df -h | grep var.opt.stardog | tr -s " " | cut -d ' ' -f 5 | tr -d "%")
```
srueg, line 78: It's probably easier to specify the target for the `df` command and you can use `--output=pcent` to only show the percentage:
```suggestion
              PERCENT_USE=$(df --output=pcent -h /var/opt/stardog | tail -n 1 | tr -d "%")
```

stardog/templates/cronjob.yaml
```diff
@@ -68,16 +68,35 @@ spec:
...
+              if [ "${PERCENT_USE}" -gt "60" ]; then echo "not enough space to test restore ${PERCENT_USE}%" ; exit 1; fi
```
srueg, line 79:
```suggestion
              if [ "${PERCENT_USE}" -gt "60" ]; then echo "not enough space to test restore: ${PERCENT_USE}% used" ; exit 1; fi
```

srueg (changes requested):

stardog/templates/cronjob.yaml
chamer81: Yes, that's cleaner.

ccremer (approved):

srueg (approved):

srueg (approved):

PR #219, opened 2020-11-23
[HAProxy] Tune Galera K8s Configuration

srueg:
Signed-off-by: Simon Rüegg <simon@rueggs.ch>

#### Checklist

- [x] Chart Version bumped


splattner (approved):

srueg:
Should we maybe factor out these common settings into a shared part which is used by all ConfigMaps?

splattner:
I had the same tough, but then was not sure if the defaults set in the common settings are best for all configs?

srueg:
Hmm right, we'd have to define defaults per config which might get quite complicated...

PR #224, opened 2020-12-16
[k8up] Upgrade K8up chart for v1.0 release

corvus-ch:

The Helm chart needs support for the rewrite of K8up which was based on the Kooper framework. One important thing to note is that Kooper took care of managing CRDs which the Operator SDK doesn't do.

#### Checklist

- [x] Squash commits
- [x] Chart Version bumped


k8up/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+# Git-treeish of the K8up archive version to fetch
+K8UP_VERSION     := master
+K8UP_CRD_VERSION := v1
+
+.PHONY: update_crds
+update_crds: ## Update the CRDs by fetching them from the k8up repository.
+	rm -rf crds
+	mkdir crds
+	curl -sL https://github.com/vshn/k8up/archive/$(K8UP_VERSION).tar.gz | tar -xj --directory crds --strip-components=5 k8up-master/config/crd/apiextensions.k8s.io/$(K8UP_CRD_VERSION)/
```
ccremer, line 16: FYI: on the next K8up release, there will be the all-in-one-YAML-file CRDs attached to the release itself (see example: https://github.com/vshn/espejo/releases/tag/v0.3.2), there will be no need to do git/tar/kustomize magic here.

The question is though, how we bundle two different versions of K8up CRDs. One is for K8s >= 1.16, the other for < 1.16 (e.g. Openshift 3.11). Either Helm detects that automatically or maybe we can/have to make use of `Capabilities.KubeVersion.Minor` (see https://helm.sh/docs/chart_template_guide/builtin_objects/) in an if-else construct.

corvus-ch: Ah, that is a nice one. As of the versioning issue: helms `crd` directory does not support templating. So I assume that `Capbabilities` is of no use here. Looking at https://helm.sh/docs/chart_best_practices/custom_resource_definitions/, we might need to go with having a separate Helm chart for the CRDs. However, I rather would like to omit them and let people install them manually or with their config management tool of choice (Project Syn in our case).

k8up/Makefile
```diff
@@ -0,0 +1,48 @@
...
+#
+# This is made phony on purpose. We do not have a local file that would trigger a rebuild.
+.PHONY: templates/crds.yaml
+templates/crds.yaml: ## Update the CRDs by fetching them from the k8up repository.
+	mkdir -p crds
+	curl -sL https://github.com/vshn/k8up/archive/$(K8UP_VERSION).tar.gz | tar -xj --directory crds --strip-components=4 k8up-master/config/crd/apiextensions.k8s.io/
+	echo -n "{{ if Capabilities.KubeVersion.Major == 1 && Capabilities.KubeVersion.Minor < 16 }}" > $@
+	# curl -sL https://github.com/vshn/k8up/releases/download/$(K8UP_VERSION)/k8up-crd-v1beta1.yaml >> $@
+	cat crds/v1beta1/* >> $@
+	echo -n "{{ else }}" >> $@
```
ccremer, line 25: Instead of a really big file with if-else clause and multiple CRD versions in it, I suggest to create separate files per CRD version still. If each file has its own if-clause in it, the effect of if-else is the same, but it makes the chart easier to understand.

k8up/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+# TODO Use the aggregated CRDs that will be attached to a release.
+# This will get rid of creating and removing the temporary `crds` folder.
+#
+# This is made phony on purpose. We do not have a local file that would trigger a rebuild.
+.PHONY: templates/crds.yaml
+templates/crds.yaml: ## Update the CRDs by fetching them from the k8up repository.
+	mkdir -p crds
+	curl -sL https://github.com/vshn/k8up/archive/$(K8UP_VERSION).tar.gz | tar -xj --directory crds --strip-components=4 k8up-master/config/crd/apiextensions.k8s.io/
+	echo -n "{{ if Capabilities.KubeVersion.Major == 1 && Capabilities.KubeVersion.Minor < 16 }}" > $@
```
ccremer, line 22: I think we should give the user more flexibility to install / maintain charts. Here, it's missing at least the a `.Values.crd.create` boolean flag:
```suggestion
	echo -n "{{ if Capabilities.KubeVersion.Major == 1 && Capabilities.KubeVersion.Minor < 16 && .Values.crd.create || .Values.crd.forceLegacyCrd }}" > $@
```

k8up/Makefile
```diff
@@ -0,0 +1,48 @@
...
+SED_CMD := gsed
+
+# TODO Use the aggregated CRDs that will be attached to a release.
+# This will get rid of creating and removing the temporary `crds` folder.
+#
+# This is made phony on purpose. We do not have a local file that would trigger a rebuild.
+.PHONY: templates/crds.yaml
+templates/crds.yaml: ## Update the CRDs by fetching them from the k8up repository.
+	mkdir -p crds
+	curl -sL https://github.com/vshn/k8up/archive/$(K8UP_VERSION).tar.gz | tar -xj --directory crds --strip-components=4 k8up-master/config/crd/apiextensions.k8s.io/
```
ccremer, line 21: Does it make sense to read the K8up version from Chart.yaml (`appVersion`) property ?

k8up/Makefile
corvus-ch: CRDs are now available at https://github.com/vshn/k8up/releases/download/v1.0.0-rc1/k8up-crd.yaml and https://github.com/vshn/k8up/releases/download/v1.0.0-rc1/k8up-crd-legacy.yaml.

ccremer:
So far the chart works and is installable (at least with default settings) in a local KIND cluster.
I've currently used the master tag/version, because rc2 still comes with too much RBAC rules and objects, which should be fixed in rc3.

ccremer:
@Kidswiss The chart is now somewhat ready for review. Can you do a review?
I haven't squashed the sommits yet, in case you're interested in the history a bit.

Kidswiss:
Installed the chart in k3d. Seems to work. Can't really comment on the code though.

srueg (approved):

ccremer:
Thanks for the Reviews. I've squashed the commits now.

This PR stays now open and waits for merge until we actually release K8up v1.0

PR #327, opened 2021-07-28
[k8up] Add "backup failed" alert based on kube_job_status_failed

bastjan:

This change allows to stop firing the alert quicker if failed jobs are cleaned.

#### Checklist

### DIFF
--- a/appuio/k8up/templates/prometheus/prometheusrule.yaml
+++ b/appuio/k8up/templates/prometheus/prometheusrule.yaml
@@ -37,7 +37,7 @@ spec:
             severity: critical
           annotations:
             summary: "K8up jobs are stuck in {{ "{{ $labels.namespace }}" }} for the last 24 hours."
-        {{- range tuple "archive" "backup" "check" "prune" "restore" }}
+        {{- range .Values.metrics.prometheusRule.jobFailedRulesFor }}
         - alert: K8up{{- . | title -}}Failed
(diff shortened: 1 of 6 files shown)


appuio/k8up/templates/prometheus/prometheusrule.yaml
```diff
@@ -43,6 +37,15 @@ spec:
             severity: critical
           annotations:
             summary: "K8up jobs are stuck in {{ "{{ $labels.namespace }}" }} for the last 24 hours."
+        {{- range tuple "archive" "backup" "check" "prune" "restore" }}
```
ccremer, line 40: I don't think we need alert rules for restores. @Kidswiss WDYT?
```suggestion
        {{- range tuple "archive" "backup" "check" "prune" }}
```

Kidswiss: Why not though? So you don't have to actively check if it failed or not. Prometheus will tell you. I'd leave it in.

ccremer: Alright.
Now that I think about it, it could be that people might want to disable those alerts for certain job types. Can this tuple made configurable @bastjan ?

Kidswiss: Making it configurable seems the most reasonable solution here.

ccremer (approved):

PR #335, opened 2021-08-18
[redis] Improve readiness check

bastjan:
Redis is seen as ready if:
- Is Master
- Is Slave and master link is up.

#### Checklist

- [ ] [DCO](https://github.com/appuio/charts/blob/master/CONTRIBUTING.md#sign-your-work) signed


appuio/redis/templates/redis-node-statefulset.yaml
```diff
@@ -164,11 +164,7 @@ spec:
               command:
                 - sh
                 - -c
-                {{- if .Values.sentinel.enabled }}
                 - /health/ping_liveness_local.sh {{ .Values.slave.livenessProbe.timeoutSeconds }}
-                {{- else }}
-                - /health/ping_liveness_local_and_master.sh {{ .Values.slave.livenessProbe.timeoutSeconds }}
-                {{- end }}
```
simu, line 171: Why do we remove this distinction between deployments which use sentinel and ones which don't?

appuio/redis/templates/redis-node-statefulset.yaml
```diff
@@ -183,11 +179,7 @@ spec:
               command:
                 - sh
                 - -c
-                {{- if .Values.sentinel.enabled }}
-                - /health/ping_readiness_local.sh {{ .Values.slave.livenessProbe.timeoutSeconds }}
-                {{- else }}
-                - /health/ping_readiness_local_and_master.sh {{ .Values.slave.livenessProbe.timeoutSeconds }}
-                {{- end }}
```
simu, line 190: Again, why are we removing the distinction between setups which use sentinel and ones which don't?

simu (approved):
LGTM overall.

Unless there's a really good reason to do so, I would avoid removing existing configuration options present in the chart (e.g. configuring different default liveness probes depending on whether sentinel is enabled or not).

In the same direction of thought: does the new readiness probe support both deployments which use sentinel and ones which don't?

Finally a side-note: I think the Lua is correct, but I'm always a bit worried when I see Lua code, since I previously got bitten badly by a typo when accessing a local variable which Lua silently interprets as "ah, you want to use this variable which I don't know, I'm just gonna return `null`".

appuio/redis/templates/redis-node-statefulset.yaml
bastjan: The template is [only ever interpreted in `sentinel` mode](https://github.com/appuio/charts/blob/e5c3053542a669df30d628aec17c23bcc4afe03b/appuio/redis/templates/redis-node-statefulset.yaml#L1). The if is superfluous.

appuio/redis/templates/redis-node-statefulset.yaml
bastjan: The template is [only ever interpreted in `sentinel` mode](https://github.com/appuio/charts/blob/e5c3053542a669df30d628aec17c23bcc4afe03b/appuio/redis/templates/redis-node-statefulset.yaml#L1). The if is superfluous.

bastjan:
Thanks for the review ✨.

> Unless there's a really good reason to do so, I would avoid removing existing configuration options present in the chart (e.g. configuring different default liveness probes depending on whether sentinel is enabled or not).

The template is [only ever interpreted in `sentinel` mode](https://github.com/appuio/charts/blob/e5c3053542a669df30d628aec17c23bcc4afe03b/appuio/redis/templates/redis-node-statefulset.yaml#L1). The if is superfluous.

> In the same direction of thought: does the new readiness probe support both deployments which use sentinel and ones which don't?

I only implemented the change for sentinel deployments. The scripts should work for both. I'll add master/slave deployments too...

> Finally a side-note: I think the Lua is correct, but I'm always a bit worried when I see Lua code, since I previously got bitten badly by a typo when accessing a local variable which Lua silently interprets as "ah, you want to use this variable which I don't know, I'm just gonna return null".

Redis runs lua in strict mode:

```
localhost:6379> eval 'return misstyped_var' 0
(error) ERR Error running script (call to f_5074e22695af46751e04867d13317d655ecfa5a2): @enable_strict_lua:15: user_script:1: Script attempted to access nonexistent global variable 'misstyped_var'
```

simu:
> Thanks for the review sparkles.
> 
> > Unless there's a really good reason to do so, I would avoid removing existing configuration options present in the chart (e.g. configuring different default liveness probes depending on whether sentinel is enabled or not).
> 
> The template is [only ever interpreted in `sentinel` mode](https://github.com/appuio/charts/blob/e5c3053542a669df30d628aec17c23bcc4afe03b/appuio/redis/templates/redis-node-statefulset.yaml#L1). The if is superfluous.
> 
> > In the same direction of thought: does the new readiness probe support both deployments which use sentinel and ones which don't?
> 
> I only implemented the change for sentinel deployments. The scripts should work for both. I'll add master/slave deployments too...
> 
> > Finally a side-note: I think the Lua is correct, but I'm always a bit worried when I see Lua code, since I previously got bitten badly by a typo when accessing a local variable which Lua silently interprets as "ah, you want to use this variable which I don't know, I'm just gonna return null".
> 
> Redis runs lua in strict mode:
> 
> ```
> localhost:6379> eval 'return misstyped_var' 0
> (error) ERR Error running script (call to f_5074e22695af46751e04867d13317d655ecfa5a2): @enable_strict_lua:15: user_script:1: Script attempted to access nonexistent global variable 'misstyped_var'
> ```

Thanks for the update, then everything LGTM from my side.

glrf (approved):
LGTM

Maybe it would make sense to put your test scripts somewhere in this repository? In a `hack/` directory or similar. It might help someone that needs to debug this chart in a year or so.

bastjan:
> Maybe it would make sense to put your test scripts somewhere in this repository? In a `hack/` directory or similar. It might help someone that needs to debug this chart in a year or so.

Good idea. I'll add them.

PR #345, opened 2021-08-31
[redis] Add option to force update the StatefulSet

glrf:
This is done using a pre-upgrade hook that deletes the initial StatefulSet with cascade=orphan (Which will only delete the Sts and not touch the pods). The helmchart will then recreate the StatefulSet.

This allows us to change fields such as PVC size which usually cannot be updated.

Signed-off-by: Fabian Fischer <fabian.fischer@vshn.ch>

#### Checklist


appuio/redis/templates/recreate.yaml
```diff
@@ -0,0 +1,76 @@
+{{- if .Values.forceUpdate }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sts-deleter
+  annotations:
+    # This is what defines this resource as a hook. Without this line, the
+    # job is considered part of the release.
```
ccremer, line 9: This should be common knowledge in Helm charts. I would remove these kind of comments.

appuio/redis/values.yaml
```diff
@@ -1006,3 +1006,8 @@ podDisruptionBudget:
   enabled: false
   minAvailable: 1
   # maxUnavailable: 1
+
+## @param forceUpdate Force update the StatefulSet.
+## If enabled the chart will recreate the StatefulSet without touching the Pods (cascade orphan), allowing you to update locked field, such as PVC size.
+##
+forceUpdate: false
```
ccremer, line 1013: I'm somewhat torn. On one hand we have helm-docs that can parse the values.yaml file for this kind of comments and then put it in the README but it requires a certain comment style. We would have to decide whether to only add this value for now, or make it for the whole document in order to be consistent (applying the boy scout rule here). I would be fine with opening an issue and accept it as technical debt though.
```suggestion
# -- forceUpdate Force update the StatefulSet.
# If enabled the chart will recreate the StatefulSet without touching the Pods (cascade orphan), allowing you to update locked field, such as PVC size.
forceUpdate: false
```

appuio/redis/templates/recreate.yaml
```diff
@@ -0,0 +1,76 @@
+{{- if .Values.forceUpdate }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sts-deleter
+  annotations:
```
ccremer, line 7: Please add the default helm labels to all the resources.

appuio/redis/templates/recreate.yaml
```diff
@@ -0,0 +1,76 @@
+{{- if .Values.forceUpdate }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: sts-deleter
```
bastjan, line 6:
```suggestion
  name: {{ printf "%s-sts-deleter" (include "common.names.fullname" .) }}
```

Always use the fullname template or setting fullNameOverride/* can lead to unexpected results.

If no additional pre/suffix is required:
```yaml
name: {{ template "common.names.fullname" . }}
```

glrf:
I switched to the `helm-docs` style comments. It's not perfect by any means, but I tried to make the helpful comments available in the README

The `values.yaml` really is a mess. Going through this I also realized, that there is a lot in this Chart that we really don't use. Maybe it would make sense to through out some of these features in the future.

appuio/redis/README.md
```diff
@@ -1,6 +1,6 @@
 # redis
```
ccremer, line 1: I forgot to mention: Please add the `README.gotmpl.md` file like the other charts do, e.g. https://raw.githubusercontent.com/appuio/charts/master/appuio/secret/README.gotmpl.md

That will alter the README slightly

ccremer (approved):

appuio/redis/values.yaml
```diff
@@ -611,62 +588,53 @@ slave:
...
+  # -- Redis(TM) Slave pod/node affinity/anti-affinity
   affinity: {}
 
-  ## Kubernetes Spread Constraints for pod assignment
+  # -- Kubernetes Spread Constraints for pod assignment
   ## ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/
   ##
   # - maxSkew: 1
   #   topologyKey: node
   #   whenUnsatisfiable: DoNotSchedule
```
ccremer, line 627: This gets rendered, looking strange
```suggestion
  ## - maxSkew: 1
  ##   topologyKey: node
  ##   whenUnsatisfiable: DoNotSchedule
```

ccremer (approved):

bastjan (approved):

PR #348, opened 2021-09-04
[stardog] Disable backup restore testing

psy-q:

The current tactic of comparing restored triple counts to backed up triple counts has caused problems with phantom databases remaining in Stardog's memory for months, leading to cluster sync problems. We're now also seeing write lock acquisition failures during the backup jobs, especially when long transactions are happening, which makes it less likely to actually get a working backup at all, let alone a restore. Backups on one large database rarely succeed even after running for 22h.

The restore also puts a lot of additional IOPS/CPU load on all nodes that could be better spent serving production workloads. The nodes not only have to restore the backup but also recompute indexes afterwards.

Thirdly, according to Stardog this measure is no longer necessary since the bug causing partial backups is said to be fixed.

We propose to create a separate, non-clustered Stardog node to test restores against instead of using the cluster. Until that's ready, this PR wants to comment out the restore functionality to see if it fixes some of the cluster sync problems.

#### Checklist

Signed-off by: Ramón Cahenzli <ramon.cahenzli@vshn.ch>

### DIFF
--- a/appuio/stardog/Chart.yaml
+++ b/appuio/stardog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: stardog
-version: 0.7.7
+version: 0.7.8
 appVersion: 7.7.2
 description: Stardog is the world’s leading knowledge graph platform for the enterprise. Stardog makes it fast and easy to turn enterprise data into knowledge.
 home: "https://www.stardog.com/"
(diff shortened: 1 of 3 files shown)


mhutter:
Instead of commenting out stuff, why not simply remove it? This is Git after all....

PR #354, opened 2021-09-10
[k8up] Update to v2.0.0

cimnine:

Updates k8up to v2.0.0

Relates to https://github.com/k8up-io/k8up/milestone/14

#### Checklist

### DIFF
--- a/appuio/k8up/Chart.yaml
+++ b/appuio/k8up/Chart.yaml
@@ -6,11 +6,10 @@ keywords:
   - backup
   - operator
   - restic
-version: 1.1.0
-appVersion: v1.2.0
+version: 1.2.0
+appVersion: v2.0.0
 sources:
   - https://github.com/vshn/k8up
-  - https://github.com/vshn/wrestic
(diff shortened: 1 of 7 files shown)


appuio/k8up/Chart.yaml
```diff
@@ -6,11 +6,10 @@ keywords:
   - backup
   - operator
   - restic
-version: 1.1.0
-appVersion: v1.2.0
+version: 1.2.0
+appVersion: v2.0.0
```
ccremer, line 10: I think due to the breaking changes in K8up usage, we should also bump this to v2 as well, so that users do not blindly upgrade seemingly minor version
```suggestion
version: 2.0.0
appVersion: v2.0.0
```

Kidswiss:
@cimnine The api groups have changed from `backup.appuio.ch` to `k8up.io`. The RBAC rules need some adjustments for that, too.

ccremer:
I've done some more work here and continued in #368

PR #360, opened 2021-09-29
[mariadb-galera] Add script preventing race condition

glrf:
#### Checklist

### DIFF
--- a/appuio/mariadb-galera/templates/recreate-script.yaml
+++ b/appuio/mariadb-galera/templates/recreate-script.yaml
@@ -1,0 +1,31 @@
+{{- if .Values.forceUpdate }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ printf "%s-sts-deleter" (include "common.names.fullname" .) }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-6"
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+data:
+  delete.sh: |
+    #!/bin/bash
+
+    NAME="{{ printf "%s" (include "common.names.fullname" .) }}"
+    NAMEPACE="{{ .Release.Namespace }}"
+
+    # Check if delete is necessary
+    FOUND=$(kubectl -n $NAMEPACE get sts $NAME -o json)
+
+    SIZE="{{ .Values.persistence.size }}"
+    FOUNDSIZE=$(echo -En $FOUND | jq -r '.spec.volumeClaimTemplates[] | select(.metadata.name=="data") | .spec.resources.requests.storage')
+
+    if [[ $FOUNDSIZE != $SIZE ]]; then
+      kubectl -n $NAMEPACE delete sts $NAME --cascade=orphan --ignore-not-found --wait=true 
+      # There is a consistency issue. It seems that it is not guaranteed that the helm controller sees the deletion before noticing the job completion.
+      # It is generally hard to find clear consistency guarantees of the Kubernetes API server.
+      # My hope is that an additional read forces the delete to be committed.
+      while kubectl -n $NAMEPACE get sts $NAME; do sleep 1; done
+      sleep 2 # Let's wait a bit to reduce the race condition likelihood if I'm wrong.
+    fi
+{{- end }}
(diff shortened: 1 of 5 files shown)


appuio/mariadb-galera/templates/recreate-script.yaml
```diff
@@ -0,0 +1,31 @@
...
+    NAME="{{ printf "%s" (include "common.names.fullname" .) }}"
+    NAMEPACE="{{ .Release.Namespace }}"
+
+    # Check if delete is necessary
+    FOUND=$(kubectl -n $NAMEPACE get sts $NAME -o json)
+
+    SIZE="{{ .Values.persistence.size }}"
+    FOUNDSIZE=$(echo -En $FOUND | jq -r '.spec.volumeClaimTemplates[] | select(.metadata.name=="data") | .spec.resources.requests.storage')
+
+    if [[ $FOUNDSIZE != $SIZE ]]; then
```
simu, line 23:
```suggestion
    if [[ "$FOUNDSIZE" != "$SIZE" ]]; then
```

appuio/mariadb-galera/templates/recreate-script.yaml
```diff
@@ -0,0 +1,31 @@
...
+      kubectl -n $NAMEPACE delete sts $NAME --cascade=orphan --ignore-not-found --wait=true
```
simu, line 24:
```suggestion
      kubectl -n "$NAMEPACE" delete sts "$NAME" --cascade=orphan --ignore-not-found --wait=true 
```

appuio/mariadb-galera/templates/recreate-script.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+    SIZE="{{ .Values.persistence.size }}"
+    FOUNDSIZE=$(echo -En $FOUND | jq -r '.spec.volumeClaimTemplates[] | select(.metadata.name=="data") | .spec.resources.requests.storage')
+
+    if [[ $FOUNDSIZE != $SIZE ]]; then
+      kubectl -n $NAMEPACE delete sts $NAME --cascade=orphan --ignore-not-found --wait=true 
+      # There is a consistency issue. It seems that it is not guaranteed that the helm controller sees the deletion before noticing the job completion.
+      # It is generally hard to find clear consistency guarantees of the Kubernetes API server.
+      # My hope is that an additional read forces the delete to be committed.
+      while kubectl -n $NAMEPACE get sts $NAME; do sleep 1; done
```
simu, line 28:
```suggestion
      while kubectl -n "$NAMEPACE" get sts "$NAME"; do sleep 1; done
```

appuio/mariadb-galera/templates/recreate-script.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: {{ printf "%s-sts-deleter" (include "common.names.fullname" .) }}
+  annotations:
+    "helm.sh/hook": pre-upgrade
+    "helm.sh/hook-weight": "-6"
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+data:
+  delete.sh: |
+    #!/bin/bash
+
+    NAME="{{ printf "%s" (include "common.names.fullname" .) }}"
```
simu, line 14: If you want more useful output from shellcheck, you should consider using lowercase variable names for variables which are not provided to the script through the environment.

With lowercase variables, you get the following for a typo of a variable which is used elsewhere:

```
Line 7:
foundsize=$(echo -En "$fonud" | jq -r '.spec.volumeClaimTemplates[] | select(.metadata.name=="data") | .spec.resources.requests.storage')
                      ^-- SC2154: fonud is referenced but not assigned.
```

while with uppercase variables you get no output for the same typo.

Note that you always get the following warning if you misspell a variable which you only use in a single place

```
Line 5:
FOUND=$(kubectl -n "$namespace" get sts $name -o json)
^-- SC2034: FOUND appears unused. Verify use (or export if used externally).
```

simu (changes requested):
Logic LGTM.

As always when writing shell scripts: Use `set -eo pipefail` and shellcheck is your best friend and helps you avoid all the subtly unintuitive behavior in shell which you're aware of but don't think about in the moment.

glrf:
Thank's for the input. As you can tell I don't have much experience in writing shell scripts.

simu (approved):

PR #374, opened 2021-11-15
Configure changelog generator

ccremer:

* Adds a changelog generator as inspired by https://github.com/ccremer/charts/releases
* Updates README

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,4 +26,4 @@ jobs:
           charts_dir: appuio
           charts_repo_url: https://charts.appuio.ch
         env:
-          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_TOKEN: "${{ secrets.RELEASE_PAT }}"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,79 +2,9 @@
 
 The APPUiO Charts project accepts contributions via GitHub pull requests. This document outlines the process to help get your contribution accepted.
 
-## Sign Your Work
-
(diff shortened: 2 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -2,79 +2,9 @@
 
 The APPUiO Charts project accepts contributions via GitHub pull requests. This document outlines the process to help get your contribution accepted.
 
-## Sign Your Work
```
simu, line 5: Why are we removing this section?

.github/workflows/release.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           charts_dir: appuio
           charts_repo_url: https://charts.appuio.ch
         env:
-          CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_TOKEN: "${{ secrets.RELEASE_PAT }}"
```
simu, line 29: What's this secret? Doesn't seem to be configured in the repo

simu (approved):
LGTM overall.

.github/workflows/release.yml
ccremer: You're right. I first need someone who has the vshnbot credentials :)
The GitHub token is insufficient as it doesn't trigger the changelog workflow ("recursive" workflow)

CONTRIBUTING.md
ccremer: We discussed a few months ago that we don't require DCO anymore, so I removed it for cleanup reasons.

bastjan (approved):
I'm happy that we're removing the DCO.

glrf (approved):

PR #382, opened 2022-01-10
New configuration syntax for Secrets

ccremer:

* Updates the generic `Secret` chart with a new value usage with additional fields.
* Added upgrade documentation
* Now also supports templated strings via `stringDataTemplates`
* Supports new value syntax (example):
```yaml
secrets:
  database: # this is the secret name
    nameTemplate: '{{ .Release.Name }}-suffix' # this allows referencing secret names from other charts
    stringData:
      my-user: example
      my-file: |
        [section]
        config = value
    data:
      my-key: some-base64-encoded-string
    labels:
      app: database
    annotations:
      my-annotation: a string value
    stringDataTemplates:
      my-template: |
        date={{ now }}
      my-name: '{{ .Values.secrets.database.name }}'
```

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`

### DIFF
--- a/appuio/secret/templates/secret.yaml
+++ b/appuio/secret/templates/secret.yaml
@@ -1,15 +1,10 @@
+{{- if kindIs "map" .Values.secrets }}
 {{- range $name, $value := .Values.secrets }}
 ---
-{{- with $value }}
 apiVersion: v1
 kind: Secret
 type: {{ default "Opaque" .type }}
 metadata:
-  {{- if .nameTemplate }}
-  name: {{ tpl .nameTemplate $ }}
-  {{- else }}
-  name: {{ $name }}
-  {{- end }}
   labels:
     {{- include "secret.labels" $ | nindent 4 }}
     {{- with .labels }}
@@ -18,6 +13,11 @@ metadata:
   {{- with .annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- if .nameTemplate }}
+  name: {{ tpl .nameTemplate $ }}
+  {{- else }}
+  name: {{ $name }}
   {{- end }}
 {{- if .stringData }}
 stringData:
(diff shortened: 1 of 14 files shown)


appuio/secret/templates/secret.yaml
```diff
@@ -1,16 +1,34 @@
...
+{{- range $name, $value := .Values.secrets }}
 ---
+{{- with $value }}
 apiVersion: v1
 kind: Secret
 type: {{ default "Opaque" .type }}
 metadata:
-  name: {{ .name }}
+  {{- if .nameTemplate }}
+  name: {{ tpl .nameTemplate $ }}
```
corvus-ch, line 9: I was wondering about that naming of that field. It indeed is treated as a template. Where can I find details on what I can do in that template? What variables can I use?

corvus-ch (approved):
Looks good to me. I might be missing something obvious or some implied knowledge. But consider documenting what can be done with those (`…Template`) variables.

appuio/secret/templates/secret.yaml
ccremer: Helm has documentation of how templating works: https://helm.sh/docs/chart_template_guide/functions_and_pipelines/
And more extensive list of functions is here: https://helm.sh/docs/chart_template_guide/function_list/.

Obviously certain functions wouldn't make sense to be used, e.g. there's no point to provide a YAML object to a name with `toYaml`, since that would produce invalid spec for the Secret.

I'm assuming that the average Helm user is aware of these functions, but maybe it's worth adding the link.

bastjan (approved):
Generally LGTM

Is this chart used often at VSHN? The changes could be made backwards compatible without much work IMO.

ccremer:
> Is this chart used often at VSHN? The changes could be made backwards compatible without much work IMO.

No, it's not used a lot. But you're right, I didn't think of making it backwards compatible.
I've now deprecated the array notation, so that it wouldn't be a breaking change. It would display a note now though.

bastjan (approved):
I like the deprecation notice and the backwards compatibility 🚀

PR #391, opened 2022-02-09
[stardog] Make ZooKeeper session timeout configurable

psy-q:

* The ZooKeeper session timeout was previously hardcoded in values.yaml but should be configurable. A higher timeout may be required when a Stardog cluster is under load.

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`

### DIFF
--- a/appuio/stardog/Chart.yaml
+++ b/appuio/stardog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: stardog
-version: 0.8.2
+version: 0.9.0
 appVersion: 7.8.1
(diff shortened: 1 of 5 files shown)


psy-q:
Oh my :man_facepalming:  Thanks, I'll do that. Only tested the templating with the default value and of course that still works like this...

psy-q:
Fixed. Sorry, I had the commits on master first and had an off-by-one error when moving them over to the correct branch...

ZuGao:
You could try to add a simple test in go but overall it looks goot. LGTM

ZuGao (approved):
LGTM

PR #394, opened 2022-02-16
Add additional parameter to control if metrics port should be exposed

TheBigLee:
The port is exposed via dedicated service and not via the main haproxy
service.

* Short summary

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`

### DIFF
--- a/appuio/haproxy/templates/service.yaml
+++ b/appuio/haproxy/templates/service.yaml
@@ -14,7 +14,7 @@ spec:
       targetPort: {{ .Values.haproxy.frontendPort }}
       protocol: TCP
       name: frontend
-    {{- if or .Values.haproxy.redisk8s.metrics.enabled .Values.haproxy.galerak8s.metrics.enabled}}
+    {{- if and (or .Values.haproxy.redisk8s.metrics.exposeLoadbalancer .VAlues.haproxy.galerak8s.metrics.exposeLoadbalancer) (or .Values.haproxy.redisk8s.metrics.enabled .Values.haproxy.galerak8s.metrics.enabled})}
     - port: 9090
       targetPort: metrics-backend
       protocol: TCP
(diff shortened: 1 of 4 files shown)


appuio/haproxy/templates/service.yaml
```diff
@@ -14,16 +14,10 @@ spec:
       targetPort: {{ .Values.haproxy.frontendPort }}
       protocol: TCP
       name: frontend
-    {{- if or .Values.haproxy.redisk8s.metrics.enabled .Values.haproxy.galerak8s.metrics.enabled}}
```
anothertobi, line 17: Could you make this configurable? It would make sense to allow exposing this through the service conditionally.

anothertobi (changes requested):
Further, please amend/add the first part of the commit message so that it says what the commit does and not why.

appuio/haproxy/templates/service.yaml
TheBigLee: Added a new parameter to make this configurable

appuio/haproxy/values.yaml
```diff
@@ -85,6 +85,8 @@ haproxy:
     nodeCount: 3
     metrics:
       enabled: false
+      exposeLoadbalancer: false
```
anothertobi, line 88: Since this used to be enabled by default, when enabling metrics, I would have the default set to true to make this a non-breaking change.

appuio/haproxy/values.yaml
```diff
@@ -100,6 +102,7 @@ haproxy:
...
```
anothertobi, line 105: Same as the other default regarding making this a non-breaking change.

anothertobi (changes requested):

anothertobi (approved):

PR #396, opened 2022-02-24
Overhaul metrics exposure for redis and mariadb

TheBigLee:
Signed-off-by: Nicolas Bigler <nicolas.bigler@vshn.ch>

* Added metrics support for the `galera` haproxy config
* Change endpoint format:
Instead of exposing the individual pods via the `path` of the url, we
will use the `Host` header and use a subdomains for route to the correct
pod.

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`


ccremer (approved):
Change LGTM, but please provide a summary of the changes in the PR description, otherwise one would need to read the code in order to understand what "overhaul" means.

PR #445, opened 2022-11-09
Ensure Stardog alert rules are picked up by Syn Prometheus

psy-q:

* Fixed Stardog alerts not being picked up by Prometheus instances deployed using the Project Syn Prometheus component

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - uses: azure/setup-helm@v1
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -11,7 +11,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - uses: azure/setup-helm@v1
+      - uses: azure/setup-helm@v3
```
psy-q, line 14: I had to upgrade the action as tests broke on the previous one.

PR #446, opened 2022-11-10
Fix tests by upgrading azure/setup-helm

psy-q:

* azure/setup-helm@v1 seemingly used Helm v2 whereas the tests expect Helm v3

This only affects GitHub actions, nothing was changed in the chart.

#### Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - uses: azure/setup-helm@v1
+      - uses: azure/setup-helm@v3
+        with:
+          version: '3.9.0'
+        id: install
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -11,7 +11,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - uses: azure/setup-helm@v1
+      - uses: azure/setup-helm@v3
+        with:
+          version: '3.9.0'
+        id: install
```
ccremer, line 17: Actually, GitHub has nowadays Helm v3 preinstalled (currently 3.10.1). So we can completely remove this action
```suggestion
```

ccremer (approved):
we can remove it, otherwise lgtm

PR #448, opened 2022-12-06
Allow specifying extra PrometheusRule labels

psy-q:

* Allow specifying extra labels on PrometheusRules (e.g. to be picked up by specific ServiceMonitors)

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`


ZuGao (commented):
Does it make sense to have the same prometheus rules for zookpeer and stardog? I haven't checked that in detail but I feel that the rules may be different for these 2 applications.

psy-q:
> Does it make sense to have the same prometheus rules for zookpeer and stardog? I haven't checked that in detail but I feel that the rules may be different for these 2 applications.

Rules no, but labels, maybe? What I'm trying to do is get VSHN's Syn-specific labels into the PrometheusRules because as they are, they're not being picked up by any ServiceMonitor and thus not monitored.

I can split it into `metrics.zookeeperRules.extraLabels` and `metrics.stardogRules.extraLabels` or something similar?

psy-q:
I've split up the rule labels now into `metrics.zookeeperRules.extraLabels` and `metrics.stardogRules.extraLabels`.

ZuGao (approved):
LGTM

PR #451, opened 2022-12-19
feat(signalilo): add ingress support

pree:

* adds ingress support to the signalilo chart

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`


simu (commented):
Can you please rebase the PR on the latest master, that should fix the failing CI check.

Also, please note that we only recommend running Signalilo in the same cluster/network as the Alertmanager instance sending alerts to Signalilo. We've never tested a setup where Signalilo is not in the same local network as the Alertmanager instance connected to it.

pree:
Thanks for the review @simu. I've added a note inside the `values.yaml`. Do you think this is sufficient, or should I add an disclaimer somewhere else?

simu (approved):
> Thanks for the review @simu. I've added a note inside the `values.yaml`. Do you think this is sufficient, or should I add an disclaimer somewhere else?

If you wouldn't mind, I'd like to also have the disclaimer in the README for the chart, other than that this LGTM now.

pree:
Added also to the README of the Signalilo chart :)

appuio/signalilo/README.md
```diff
@@ -18,6 +18,11 @@ The README.md file is automatically generated with helm-docs!
 Edit the README.gotmpl.md template instead.
 -->
 
+## Deployment considerations
+
+It is recommended to deploy Signalilo only in the same cluster or network as the alertmanager instance, which is sending the alerts.
```
simu, line 23:
```suggestion
It is recommended to deploy Signalilo only in the same cluster or network as the alertmanager instance which is sending the alerts.
```

simu:
Thanks for the contribution :+1:

PR #458, opened 2023-02-09
Add option to expose a filterproxy

glrf:
With this we can forward a view of the https://github.com/vshn/exporter-filterproxy. Is used to provide the metrics related to the namespace.

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`

### DIFF
--- a/appuio/haproxy/README.gotmpl.md
+++ b/appuio/haproxy/README.gotmpl.md
@@ -138,19 +138,33 @@ Set `haproxy.config` to `redisk8s` to use the Redis configuration with DNS servi
 
 | Parameter              | Description            | Default
 |---                     | ---                    | ---
-| `filterproxy.enabled` | Enables exposing an [exporter-filterproxy](https://github.com/vshn/exporter-filterproxy) that will be filtered based on the namespace | `false`
-| `filterproxy.url` | The URL of the exporter-filterproxy to expose | `""`
+| `haproxy.filterproxy.enabled` | Enables exposing an [exporter-filterproxy](https://github.com/vshn/exporter-filterproxy) that will be filtered based on the namespace | `false`
+| `haproxy.filterproxy.url` | The URL of the exporter-filterproxy to expose | `""`
 
 ## Galera and Redis metrics
 
 By setting the `haproxy.galerak8s.metrics.enabled` or `haproxy.redisk8s.metrics.enabled` parameters, you can expose the metrics of the backend pods.
 
-For Galera the endpoints are:
+For Galera the endpoints are either:
 * `<haproxy_service_ip>:9090/mariadb/0`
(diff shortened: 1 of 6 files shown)


simu (approved):
LGTM overall

TheBigLee (approved):
LGTM

appuio/haproxy/README.gotmpl.md
```diff
@@ -134,17 +134,23 @@ Set `haproxy.config` to `redisk8s` to use the Redis configuration with DNS servi
...
-* `<haproxy_service_ip>:9090/metrics/mariadb-1/metrics`
+* `<haproxy_service_ip>:9090/mariadb/0`
+* `<haproxy_service_ip>:9090/mariadb/1`
 * `...`
 
 For Redis the endpoints are:
-* `<haproxy_service_ip>:9090/metrics/redis-0/metrics`
-* `<haproxy_service_ip>:9090/metrics/redis-1/metrics`
+* `<haproxy_service_ip>:9090/redis/0`
+* `<haproxy_service_ip>:9090/redis/1`
```
anothertobi, line 155: The new paths are correct for the filterproxy implementation. The regular metrics path would actually be `<haproxy_service_ip>:9090/metrics` with a host header of `redis-0`, `redis-1`, etc.

```
curl -H "Host: redis-0" haproxy:9090/metrics
```

anothertobi (changes requested):
My understanding is that the exporter-filterproxy would return all metrics filtered for the given namespace and not the metrics for a specific backend node.

Example:
A request to `haproxy:9090/mariadb/0` would return all metrics for the namespace the HAProxy is running in and not just the metrics of the backend `galera-node-0` which would not require any namespace restriction.

Therefore, a single endpoint (on a different port) seems more appropriate.

appuio/haproxy/README.gotmpl.md
glrf: You're right this doesn't explain the whole picture (and before the documentation was completely wrong though) I'll improve it

glrf:
> My understanding is that the exporter-filterproxy would return all metrics filtered for the given namespace and not the metrics for a specific backend node.
> 
> Example: A request to `haproxy:9090/mariadb/0` would return all metrics for the namespace the HAProxy is running in and not just the metrics of the backend `galera-node-0` which would not require any namespace restriction.

No `haproxy:9090/mariadb/0` won't be forwarded to the filterproxy at all but to the first mariadb instance. Also the filerproxy doesn't return all metrics for a namespace but can proxy specific exporters and filter them by label (in this case by namespace and in the first iteration will be used to expose kube-state-metrics only)

anothertobi (approved):
The new `/metrics` shortcut had me confused. LGTM 👍

PR #468, opened 2023-03-23
Safer PDBs (only for specific instance and replicaCount >1)

psy-q:

* Only install a PDB when the Stardog replica count is > 1. Otherwise this could prevent nodes from being drained as it refuses to shut down the only pod in a single-replica setup.
* Scope the PDB to a specific instance.

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`

### DIFF
--- a/appuio/stardog/templates/pdb.yaml
+++ b/appuio/stardog/templates/pdb.yaml
@@ -1,4 +1,4 @@
-{{- if (gt .Values.replicaCount 1) }}
+{{- if (gt (.Values.replicaCount | int) 1) }}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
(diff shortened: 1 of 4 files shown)


appuio/stardog/templates/pdb.yaml
```diff
@@ -7,3 +8,4 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/name: {{ include "stardog.name" . }}
```
zugao, line 10: All stardog instances have the same label name `app.kubernetes.io/name: stardog` which means this PDB will be created for each environment and will be applied to all instances. You might want to add `app.kubernetes.io/instance` to the matchLabels selector.

psy-q: :scream: 

Thanks, fixed now, I hope.

PR #476, opened 2023-06-13
Update codeowners for Trifid etc.

psy-q:

* Add @ludovicm67  in place of sandhose


ludovicm67 (approved):
Thanks @psy-q, looks great! 👍

PR #485, opened 2023-07-18
Support exposing a filterproxy for non k8s galera

glrf:
We also need the feature introduced in #458 for `galera` not just for `galerak8s`.

This pr simply adds the same config for `galera`.

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`


TheBigLee (approved):
LGTM

PR #494, opened 2023-09-11
Adding https checks for HAproxy and Galera

wejdross:

* This change ensures that Haproxy sends traffic only to nodes with wsred_ready=1 metric set. It queries prometheus endpoint and parses config. Buffer size for haproxy was raised from 16kB to 300kB, otherwise Haproxy can't "see" metrics. 

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`
- [ ] PR contains the label that identifies the chart, e.g. `chart/<chart-name>`
- [ ] PR contains the label that identifies the type of change, which is one of

### DIFF
--- a/appuio/haproxy/Dockerfile
+++ b/appuio/haproxy/Dockerfile
@@ -1,4 +1,4 @@
-from haproxytech/haproxy-debian:2.7.3
+from haproxytech/haproxy-debian:2.8.3
 
 RUN curl -o mysql.deb https://dev.mysql.com/get/mysql-apt-config_0.8.26-1_all.deb || exit 1 \
     dpkg -i mysql.deb || exit 1 ; \
--- a/appuio/haproxy/Dockerfile
+++ b/appuio/haproxy/Dockerfile
@@ -1,0 +1,7 @@
+from haproxytech/haproxy-debian:2.7.3
+
+RUN curl -o mysql.deb https://dev.mysql.com/get/mysql-apt-config_0.8.26-1_all.deb || exit 1 \
+    dpkg -i mysql.deb || exit 1 ; \
+    apt-get update || exit 1 ; \
+    apt-get install -y default-mysql-client || exit 1 ; \
+    rm -rf /var/lib/apt/lists/* mysql.deb ;
--- a/appuio/haproxy/README.md
+++ b/appuio/haproxy/README.md
@@ -1,6 +1,6 @@
 # haproxy
 
-![Version: 2.3.2](https://img.shields.io/badge/Version-2.3.2-informational?style=flat-square) ![AppVersion: 2.7.3](https://img.shields.io/badge/AppVersion-2.7.3-informational?style=flat-square)
+![Version: 2.4.0](https://img.shields.io/badge/Version-2.4.0-informational?style=flat-square) ![AppVersion: 2.7.3](https://img.shields.io/badge/AppVersion-2.7.3-informational?style=flat-square)
 
 A Helm chart for HAProxy which can be customized by a config map.
 
--- a/appuio/haproxy/README.md
+++ b/appuio/haproxy/README.md
@@ -1,6 +1,6 @@
 # haproxy
 
-![Version: 2.3.1](https://img.shields.io/badge/Version-2.3.1-informational?style=flat-square) ![AppVersion: 2.7.3](https://img.shields.io/badge/AppVersion-2.7.3-informational?style=flat-square)
+![Version: 2.3.2](https://img.shields.io/badge/Version-2.3.2-informational?style=flat-square) ![AppVersion: 2.7.3](https://img.shields.io/badge/AppVersion-2.7.3-informational?style=flat-square)
 
 A Helm chart for HAProxy which can be customized by a config map.
 
--- a/appuio/haproxy/files/galera-check.sh
+++ b/appuio/haproxy/files/galera-check.sh
@@ -18,7 +18,6 @@ fi
 # sample return -> "WSREP_CLUSTER_STATUS Primary WSREP_CONNECTED ON WSREP_LOCAL_STATE 4 WSREP_READY ON"
 return_val=$(/usr/bin/mysql -h $3 -u root -p"$pass" -e "SELECT VARIABLE_NAME, VARIABLE_VALUE FROM information_schema.global_status WHERE VARIABLE_NAME IN ('wsrep_cluster_status','wsrep_ready','wsrep_local_state','wsrep_connected') ORDER BY VARIABLE_NAME ASC;" -Ns)
 if [ $? -ne 0 ]; then
-  echo Error: Unable to connect to MySQL or query failed
   exit 1
 else
   to_check=$(echo $return_val | cut -d' ' -f2)
--- a/appuio/haproxy/files/galera-check.sh
+++ b/appuio/haproxy/files/galera-check.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+    
+# Run the MySQL query and store the result in a Bash variable
+
+wsrep_ready=`/bin/cat /secrets/mariadb-root-password | /usr/bin/mysql -h $3 -u root -p -e "SELECT VARIABLE_VALUE FROM information_schema.global_status WHERE VARIABLE_NAME = 'wsrep_ready';" -N`
+if [ $? -ne 0 ]; then
+  echo Error: Unable to connect to MySQL or query failed
+  exit 1
+fi
+
+if [ $wsrep_ready != "ON" ]; then
+  echo Error: wsrep_ready is not ON, actual status is: $WSREP_READY
+  exit 1
+fi
+
+exit 0 
--- a/appuio/haproxy/files/galera-check.sh
+++ b/appuio/haproxy/files/galera-check.sh
@@ -24,14 +24,14 @@ else
   to_check=$(echo $return_val | cut -d' ' -f2)
 
   if [ $to_check != "Primary" ]; then
-    echo "Error: WSREP_CLUSTER_STATUS is not 'Primary'"
+    echo "Error: WSREP_CLUSTER_STATUS is not 'Primary' it's: $to_check"
     exit 1
   fi
 
   to_check=$(echo $return_val | cut -d' ' -f4)
 
   if [ $to_check != "ON" ]; then
-    echo "Error: WSREP_CONNECTED is not 'ON'"
+    echo "Error: WSREP_CONNECTED is not 'ON' it's: $to_check"
     exit 1
   fi
 
@@ -44,7 +44,7 @@ else
 
   to_check=$(echo $return_val | cut -d' ' -f8)
   if [ $to_check != "ON" ]; then
-    echo "Error: WSREP_READY is not 'ON'"
+    echo "Error: WSREP_READY is not 'ON' it's: $to_check"
     exit 1
   fi
 fi
--- a/appuio/haproxy/files/galera-check.sh
+++ b/appuio/haproxy/files/galera-check.sh
@@ -1,8 +1,13 @@
 #!/bin/bash
-    
-### This script is used to check the status of the Galera cluster
+
+#====================================================================================================
+### This script is used to check the status of the Galera cluster by HAProxy
 ### Maintainer: https://github.com/wejdross 
-### It takes hash of checks and expected values as input and returns 0 if all checks are successful, 1 otherwise 
+### It's sh/dash compatible, because it's used in a haproxy container and bash shell has serious issues to spawn 
+### logic is simple, it takes 4 parameters and check if they are equal to the expected value
+### parameters are returned always in the same manner thanks to ORDER BY, if any of the parameters is not equal to the expected value, script will exit with 1
+#====================================================================================================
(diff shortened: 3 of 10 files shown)


TheBigLee (commented):
I thought about that approach for a while now, and I don't think this is a good idea to use the metrics exporter for this:
* The result from the exporter can get "huge" for really big instances with loads of databases and tables and the `300k` in `tune.bufsize` might not be enough.
* It puts a lot of strain on the haproxy if he has to parse the result every time for that `mysql_global_status_wsrep_ready` string. Especially for big clusters with loads of metrics
* It puts unnecessary strain on the cluster as it queries the endpoint every second.
* If the exporter fails (or the query is too slow or too big), the whole instance goes down.

wejdross:
so what do You suggest? should we write custom sidecar? how is that solution better?

TheBigLee:
> so what do You suggest? should we write custom sidecar? how is that solution better?
Yes and using a script to determine if the galera node is up and running.

Also using `mysql_global_status_wsrep_ready` might not be enough to determine if a node is ready to accept connections. Afaik this value is true as well if the node is in the DONOR state. However, the node is blocked during that state and can't accept queries. Which again will result in downtimes as haproxy assumes the node is up.

wejdross:
I took info from docs:
```
Whether the server is ready to accept queries. If this status is OFF, almost all of the queries will fail with:
ERROR 1047 (08S01) Unknown Command
```

Ready to receive queries sounds like it should be ready to receive queries :D 
But if You don't like this solution, what do You suggest? custom sidecar parsing metrics and exposing http endpoint with statuses 200/503? 

the only reason why I like solution with sidecar is this buffer which is hard to estimate and 244kB is not that small amout. but saving memory in haproxy would cause that we need to run 3 sidecars on each database engine consuming resources as well. 

It's Your call @TheBigLee I'm OK with both solutions

zugao:
Can [this](https://www.haproxy.com/documentation/hapee/latest/service-reliability/health-checking/mysql-health-checks) be some kind of an alternative solution ?

wejdross:
> Can [this](https://www.haproxy.com/documentation/hapee/latest/service-reliability/health-checking/mysql-health-checks) be some kind of an alternative solution ?

unfortunately not, whole thing is to detect when galera is ready, not mariadb itself :/

zugao:
> > Can [this](https://www.haproxy.com/documentation/hapee/latest/service-reliability/health-checking/mysql-health-checks) be some kind of an alternative solution ?
> 
> unfortunately not, whole thing is to detect when galera is ready, not mariadb itself :/

But MariaDB will not respond ok if galera cluster is not working on that node no?

TheBigLee:
> > > Can [this](https://www.haproxy.com/documentation/hapee/latest/service-reliability/health-checking/mysql-health-checks) be some kind of an alternative solution ?
> > 
> > 
> > unfortunately not, whole thing is to detect when galera is ready, not mariadb itself :/
> 
> But MariaDB will not respond ok if galera cluster is not working on that node no?

It can actually. MariaDB can respond ok but not accept queries (eg. when in DONOR mode)
Furthermore, this is a haproxy enterprise feature.

zugao:
Unfortunately I have limited knowledge with reverse proxies / load balancers such as HAProxy and Nginx. Nevertheless I took some time and read relevant docs and some blogs. There are 2 facts:

1) HAProxy supports running a scripts to check for various galera nodes.
2) Increasing the [tune.bufsize](https://cbonte.github.io/haproxy-dconv/2.2/configuration.html#3.2-tune.bufsize) is strongly not recommended.

Now my 2 cents on these:
1) Since this approach does not imply changing shady parameters I definitely would go with it, especially in our case where we might have huge databases. Now how we want to do it - via a simple script or full blown web server, we can discuss and decide as a team.
2) From the docs it seems that changing the value of this parameter implies that other parameters have to be adjusted as well, otherwise nasty things may happen. My main concern is that later on we might have hidden issues because of this parameter which could take us an unbelievable amount of time to figure out the cause. 

Conclusion: 
I don't see this improvement as a serious reason to change a controversial parameter with `20x` the default value, the risk is not justified.

@Kidswiss we might also want to hear from you on this sensitive topic.

Kidswiss:
I agree with the consent here, given that the output of the exporter is so huge and we need to adjust not recommended settings, we should look for an alternative.

IMHO a script check would be the optimal solution here, the script could connect to the instance, query the wsrep status (or other relevant things, like donor status) and return back the readiness of a given node.

I would not go with webserver that exports that information again. Then we'll have yet another long running thing that could break. Also the indirect nature of that approach would probably introduce new edge cases.

wejdross:
Short summary:
- extended script logic to galera and galerak8s
- added Dockerfile that builds haproxy image with `mysql` command required to our check
- tested it locally with `helm template` and it looks good
- tested logic in dev cluster by editing deployments, not by installing new ones

I'd be very happy to get some critic as I'm very inexperienced in helm charts development and I don't even know what should I look for.

appuio/haproxy/templates/configmap-galera-checkscript.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if or (eq .Values.haproxy.config "galera") (eq .Values.haproxy.config "galerak8s") }}
+
+kind: ConfigMap
+apiVersion: v1
+data:
+  script.sh: "#!/bin/bash\n\n# Run the MySQL query and store the result in a Bash
```
Kidswiss, line 6: Can't helm read from a file and insert the string here?

Having the script separate in a readable form would greatly help the reviewability/maintainability of this script.

Kidswiss (approved):
I'm also no helm expert.

Just one nit, the rest looks good.

appuio/haproxy/templates/configmap-galera-checkscript.yaml
TheBigLee: +1 on that. I would try to do this via file if possible.

appuio/haproxy/templates/configmap-galera.yaml
```diff
@@ -37,10 +39,12 @@ data:
...
       mode tcp
       option srvtcpka
+      option external-check
+      external-check command /var/lib/haproxy/script.sh
       balance {{ $galera.balance }}
       {{ if and $galera.check.enabled $galera.check.mysql.enabled }}option mysql-check user {{ $galera.check.mysql.user }}{{ end }}
       default-server init-addr none {{ if $galera.check.enabled }}check {{ end }}resolvers mydns
       {{- range $index, $node := .Values.haproxy.galera.nodes }}
-      server node-{{ $index }} {{ $node.address }}:{{ default "3306" $node.port }}{{ if $node.backup }} backup{{ end }}
+      server node-{{ $index }} {{ $node.address }}:{{ default "3306" $node.port }} {{ if $node.backup }} backup{{ end }}
```
TheBigLee, line 48: double whitespace here

appuio/haproxy/templates/configmap-galera-checkscript.yaml
```diff
@@ -0,0 +1,13 @@
...
+    variable\nwsrep_ready=$(/bin/cat /secrets/mariadb-root-password | /usr/bin/mysql
+    -h $3 -u root -p -e \"SELECT VARIABLE_VALUE FROM information_schema.global_status
```
TheBigLee, line 8: Instead of using `cat`you can mount the secret as env variable and pass it to the mysql command like this:
```
mysql -h $3 -u root -p $MARIADB_ROOT_PASSWORD -e [...]
```

appuio/haproxy/templates/deployment.yaml
```diff
@@ -71,13 +71,26 @@ spec:
           volumeMounts:
             - name: haproxy-config
               mountPath: /etc/haproxy/
+            - mountPath: /var/lib/haproxy
+              name: haproxy-script
+            - mountPath: /secrets
```
TheBigLee, line 76: Mount the password as env variable instead of a volume, to have the root password as env variable

appuio/haproxy/values.yaml
```diff
@@ -37,7 +37,7 @@ service:
   additionalPorts: []
 
 haproxy:
-  config: ldap-tls
+  config: galerak8s
```
TheBigLee, line 40: don't change the default config here. Leave it at `ldap-tls`

appuio/haproxy/Dockerfile
```diff
@@ -0,0 +1,7 @@
+from haproxytech/haproxy-debian:2.7.3
```
TheBigLee, line 1: Please use the latest version (currently 2.8.3)

TheBigLee:
The script doesn't seem to work in the current state:
Haproxy logs the following error:
```
Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: Enter password: ERROR 2002 (HY000): Can't connect to MySQL server on '11.33.187.2' (115)
[WARNING]  (1) : Process 21639 exited with code 1 (Exit)
```

appuio/haproxy/templates/configmap-galera-checkscript.yaml
wejdross: because script does not have access to env variables, therefore it must be attached as file :/ script is run as subprocess of haproxy and haproxy does not pass envs into it

appuio/haproxy/templates/deployment.yaml
wejdross: because script does not have access to env variables, therefore it must be attached as file :/ script is run as subprocess of haproxy and haproxy does not pass envs into it

appuio/haproxy/Dockerfile
wejdross: done, also pushed

appuio/haproxy/values.yaml
wejdross: restored

appuio/haproxy/templates/configmap-galera-checkscript.yaml
wejdross: As a compromise I fixed formatting - wdyt?

Kidswiss: A separate file would make it easier to check with shellcheck.

There's a way to read files in helm: https://helm.sh/docs/chart_template_guide/accessing_files/

Something like this should do the trick:
```yaml
data:
  script.sh: |-
        {{ .Files.Get "script.sh" }}
```

appuio/haproxy/README.md
```diff
@@ -1,6 +1,6 @@
 # haproxy
 
-![Version: 2.3.1](https://img.shields.io/badge/Version-2.3.1-informational?style=flat-square) ![AppVersion: 2.7.3](https://img.shields.io/badge/AppVersion-2.7.3-informational?style=flat-square)
+![Version: 2.3.2](https://img.shields.io/badge/Version-2.3.2-informational?style=flat-square) ![AppVersion: 2.7.3](https://img.shields.io/badge/AppVersion-2.7.3-informational?style=flat-square)
```
zugao, line 3: I would set the version of the helm chart to `2.4.0`.

appuio/haproxy/files/galera-check.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+    
+# Run the MySQL query and store the result in a Bash variable
+
+wsrep_ready=`/bin/cat /secrets/mariadb-root-password | /usr/bin/mysql -h $3 -u root -p -e "SELECT VARIABLE_VALUE FROM information_schema.global_status WHERE VARIABLE_NAME = 'wsrep_ready';" -N`
```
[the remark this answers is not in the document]
TheBigLee: Furthermore as discussed, please check all of the following variables:
```
wsrep_local_state: 4 # see https://galeracluster.com/library/documentation/node-states.html#node-state-changes
wsrep_cluster_status: Primary     
wsrep_connected: ON

**Comment by TheBigLee** [comment]
Please run your script through [shellcheck](https://www.shellcheck.net/) when the script is finished to catch some bash specific issues.

File: appuio/haproxy/files/galera-check.sh [hunk 9]
```diff
@@ -0,0 +1,51 @@
...
+export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+pass=$(/bin/cat /secrets/mariadb-root-password)
+
+if [ -z "$pass" ]; then
+  echo "Error: password is empty, please check mount and permission declared in haproxy deployment"
+  exit 1
+fi
+
+# sample return -> "WSREP_CLUSTER_STATUS Primary WSREP_CONNECTED ON WSREP_LOCAL_STATE 4 WSREP_READY ON"
+return_val=$(/usr/bin/mysql -h $3 -u root -p"$pass" -e "SELECT VARIABLE_NAME, VARIABLE_VALUE FROM information_schema.global_status WHERE VARIABLE_NAME IN ('wsrep_cluster_status','wsrep_ready','wsrep_local_state','wsrep_connected') ORDER BY VARIABLE_NAME ASC;" -Ns)
```
TheBigLee, line 19: I would redirect stderr to /dev/null to avoid unessesary output in the logs

PR #496, opened 2023-10-26
Adjust health check to return 503 if all backends are down

TheBigLee:

If all backends are down, the haproxy instance should not be served anymore by kubernetes and the readiness probe should fail. Furthermore, we also fail the liveness probe to enforce a restart of haproxy, as this can resolve the issue in some cases (eg. network or dns issues).

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`


appuio/haproxy/haproxy-2.4.0.tgz
wejdross, line 1: is this necessary?

appuio/haproxy/spks.values
wejdross, line 1: is this necessary?

wejdross (approved):
LGTM

zugao (approved):
LGTM

PR #499, opened 2024-01-18
Add template for HttpCheck alert configuration

martini-source:

* Adds a helm template to configure the 'for' key of stardog HttpCheck alerts.

#### Checklist

- [x] Chart Version bumped
- [x] I have run `make docs`

### DIFF
--- a/appuio/stardog/templates/monitoring/stardog-rules.yaml
+++ b/appuio/stardog/templates/monitoring/stardog-rules.yaml
@@ -48,7 +48,7 @@ spec:
     {{- if .Values.ingress.enabled }}
     - alert: HttpCheck
       expr: probe_success{instance="https://{{ .Values.ingress.host }}/admin/healthcheck"} != 1
-      for: {{ (((.Values.alerts).httpCheck).for) | default 1m | quote }}
+      for: {{ (((.Values.alerts).httpCheck).for) | default 1m }}
       annotations:
         summary: Http check failed
         description: "Http check failed on {{ "{{" }} $labels.instance {{ "}}" }}"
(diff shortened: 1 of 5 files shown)


appuio/stardog/templates/monitoring/stardog-rules.yaml
```diff
@@ -48,7 +48,7 @@ spec:
     {{- if .Values.ingress.enabled }}
     - alert: HttpCheck
       expr: probe_success{instance="https://{{ .Values.ingress.host }}/admin/healthcheck"} != 1
-      for: 1m
+      for: "{{ "{{ if (((.Values.alerts).httpCheck).for) }}" }}{{ .Values.alerts.httpCheck.for }}{{ "{{ else }}" }}1m{{ "{{ end }}" }}"
```
davidgubler, line 51: I'm not sure but wouldn't something simpler like this work?

{{ default "1m" .Values.alerts.httpCheck.for | quote }}

appuio/stardog/values.yaml
```diff
@@ -92,7 +92,9 @@ metrics:
     repository: sscaling/jmx-prometheus-exporter
     tag: 0.12.0
     pullPolicy: IfNotPresent
-
+alerts:
```
davidgubler, line 95: This will apply the "30m" value to all environments that use this helm chart. I don't think this is what we want?

This value should only be set in the repo which invokes this chart.

davidgubler:
Merging all the commits into one would also be nice.

appuio/stardog/templates/monitoring/stardog-rules.yaml
martini-source: I did try this at first see bc991eebfa2c321ef4da54511da9298d7422aac1, but the tests kept failing with a parse error for "1m", written with and without quotes as well as with pipe to quotes and without.

davidgubler: Hmmm I think you should try the way I've proposed (I've used that for rocketchat-deploy therefore I'm pretty sure it works). I'm not convinced that the way you tried does the same thing.

appuio/stardog/values.yaml
martini-source: That's true, done.

martini-source:
I figured I would squash all into one when actually merging.

PR #520, opened 2025-04-23
Fix recurring errors when using TLS/SSL

Kidswiss:
Turns out haproxy needs to be told that the connection is TLS/SSL twice. Once for the `tcp-check` connect and then for the `server-template`.

If `ssl` is not specified in the `server-template` config, then there will be some plaintext tcp connections. These seem to be separate from the `tcp-check` connections though.

Redis will print an error such as this when it's running in TLS/SSL mode and a plaintext connection gets opened:

```
1:M 14 Apr 2025 14:41:03.686 # Error accepting a client connection: error:0A00010B:SSL routines::wrong version number (addr=11.35.244.11:34886 laddr=11.35.244.4:6379)
```

They are mostly annoying, but are harmless.

* Short summary

#### Checklist

- [ ] Chart Version bumped
- [ ] I have run `make docs`
- [ ] PR contains the label that identifies the chart, e.g. `chart/<chart-name>`
- [ ] PR contains the label that identifies the type of change, which is one of


zugao (approved):
LGTM

PR #562, opened 2026-09-04
Add an index.html pointing at index.yaml

arska:
## Why

A Helm repository's only contract is `index.yaml`. Nothing ever requests `/`, so what `/` returns has never said anything about whether the repository works.

That is not hypothetical. An external link audit of the VSHN estate classified every hostname by what `/` returned, and read this host by that answer alone. The repository was serving its index correctly the entire time.

## What this adds

An `index.html` on `gh-pages` showing:

- the `helm repo add` / `update` / `search` commands
- how to install a chart from the repository
- links to the chart sources, the raw `index.yaml`, and the product documentation

Someone who types the hostname now learns what it is and how to use it.

## Not a redirect

An earlier revision of this PR redirected to `index.yaml`. Changed on review: sending a person to raw YAML proves the repository is alive but tells them less than the command they came for. Helm is unaffected either way, since it never requests `/`.

## Why it is safe on a generated branch

`gh-pages` is written by `helm/chart-releaser-action`, which **adds** packages and updates `index.yaml` but never cleans the branch. `CNAME` and `README.md` have survived there across many releases on exactly those terms, which is the evidence rather than the assumption.

The file carries an HTML comment stating it is hand-maintained and why, so the next person does not assume it is generated output and delete it.

## Verifying

After merge, `/` should render the page, and `helm repo add` should be unchanged. Rendered locally before pushing rather than checked by reading the markup.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,78 @@
+<!DOCTYPE html>
+<!--
+  Hand-maintained on gh-pages, not generated. Do not delete as stale build output.
+
+  This branch is written by helm/chart-releaser-action, which adds packages and
+  updates index.yaml but never cleans the branch, so this file survives releases.
+  CNAME and README.md have survived on the same terms.
+
+  It exists because a Helm repository's only contract is index.yaml. Nothing
+  requests "/", so "/" said nothing about whether the repository worked, and an
+  external audit read this host by that answer alone: it 404s at "/" while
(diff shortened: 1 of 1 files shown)


arska:
@BacLuc any possible objection to putting a small page on https://charts.appuio.ch/, so it doesn't return a 404 error? My VSHN-wide link checker flagged it :)

